PR #39, opened 2021-09-09
Handle identity failure

### LINKED ISSUE #29: After sessionID is sent via failure dialog, clicking on failed account crashes the app.
**Description**
Hashed sessionID functionality: On account page, after sessionID is sent via failure dialog, clicking on failed account crashes the app.

Steps to Reproduce:
* in failure dialog on account page (not identity page), “contact support”
* sent support e-mail with sessionID to NB
* initial account marked as failed
* clicking on failed account crashes the app

**Expected Result**
No crash.

**Actual Result**
Crash.

**Versions**
* Concordium Testnet Wallet v1.0(14)
* iPhone 11, iOS 14.5.1

### LINKED ISSUE #30: Leaving app while waiting for id issuing (to fail) and then returning on account page crashes the app.
**Description**
Hashed sessionID functionality: Leaving app while waiting for id issuing (to fail) and then returning on account page crashes the app.

Steps to Reproduce:
* issued fake identity
* went to account page
* navigated away from app for a while
* came back to app on account page
* failure dlg was briefly shown
* app crashed 

**Expected Result**
No crash.

**Actual Result**
Crash.

**Versions**
* Concordium Testnet Wallet v1.0(14)
* iPhone 11, iOS 14.5.1

### LINKED ISSUE #36: Improve dlg shown on failed identity
**Task description**

For iOS, this issue is a merger between #26, #27 and #28, and it might also extend #29 and #30, as these bugs are probably related to this.

The dialog shown on failed identities that has already been somewhat implemented, needs a few improvements before release. The full flow on how it should work has been sketched out in Figma, and it's called "Flow 18 - Failed identity issuance".

The main differences from what has already been implemented, is the addition of a "Cancel" button on the dialog shown when the identity / initial account fails. The view for failed identity cards has also been updated to show the hashed session ID, so the users can come back to find it later. This change also means, that the user cannot always be sent back to the intro flow, as that would make it impossible to go a find the hashed session ID again. The Figma flow should explain how the new routing should work.

kristiyandobrev:
## Purpose

Based on #36, this PR is to introduce handling of the identity failure. Besides that it Introduces three new utility structs `CopyPasteHelper`, `MailHelper` and `HashingHelper`. 

## Checklist

- [x] The code compiles without warnings.
      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG.

Closes #36
Closes #30  
Closes #29


ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift
```diff
@@ -223,26 +223,22 @@ class AccountsPresenter: AccountsPresenterProtocol {
...
+        let identities = dependencyProvider.storageManager().getIdentities()
+        let failedIdentities = identities.filter { $0.state == .failed }
+        
+        for identity in failedIdentities {
+            guard let reference = identity.hashedIpStatusUrl else {
+                return
+            }
+            
+            // if there is an account associated with the identity, we delete the account and show the error
+            if let account = dependencyProvider.storageManager().getAccounts(for: identity).first {
```
abizjak, line 235: Why is the first account the correct one to remove?

I suppose if everything works there should not be more than one account if the identity is failed?

ConcordiumWallet/Views/Identities/IdentityList/IdentitiesPresenter.swift
```diff
@@ -57,26 +57,21 @@ class IdentitiesPresenter: IdentityGeneralPresenter {
...
-                        self.delegate?.tryAgainIdentity()
-                        
-                    })
-                    break // we break here because if there are more accounts that failed, we want to show that later on
+        let failedIdentities = identities.filter { $0.state == .failed }
+
+        for identity in failedIdentities {
+            guard let reference = identity.hashedIpStatusUrl else {
+                return
+            }
```
abizjak, line 65: Why is this a sensible thing to do? Why are all remaining failed identities skipped if one of them does not have a URL?

kristiyandobrev: Good catch. It needs `continue` instead of `return`

ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift
kristiyandobrev: To be fair, I am not sure. I have seen this approach other places (in the project) as well and I have had the same question towards @jens-concordium. But I was told that removing the first associated account is just fine.

abizjak: Ok. I think this kind of handling is very fiddly and error prone, and it would be better to have a special "pending initial account" field/status. But that is a significant refactoring as far as I can see, so let's go with this.

ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift
```diff
@@ -223,26 +223,22 @@ class AccountsPresenter: AccountsPresenterProtocol {
...
-                        self.dependencyProvider.storageManager().removeIdentity(identity)
-                        self.delegate?.tryAgainIdentity()
-                    })
-                    break // We break here because if there are more accounts that failed, we want to show that later on
+        let identities = dependencyProvider.storageManager().getIdentities()
+        let failedIdentities = identities.filter { $0.state == .failed }
+        
+        for identity in failedIdentities {
+            guard let reference = identity.hashedIpStatusUrl else {
+                return
```
abizjak, line 231: I guess this should also be changed to `continue` as in the other place?

ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift
jens-concordium: The user cannot create more accounts with pending or failed account, so I guess from my perspective it made sense to just remove the first one. In any case, if an identity fails it should not have any accounts in the wallet.

ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift
kristiyandobrev: Correct. Changed 😊

ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift
abizjak: I understand that in principle it should work, it is just a bit dirty :) In any case, it is used consistently so lets leave this.

abizjak (approved):

PR #40, opened 2021-09-10
Sort address book alphabetically

### LINKED ISSUE #19: Sort address book entries alphabetically
**Description**
The desktop wallet and the Android version of the mobile wallet sorts address book entries alphabetically. It would be nice if the iOS app does the same.

**Steps to Reproduce**
Watch entries in the address book on iOS and notice that they are sorted in the order they are added.

**Expected Result**
They are soreted 0-9 and then A-Z.

**Actual Result**
They are sorted in the order they are added.

**Versions**
- Software Version: Testnet 0.1.68
- OS: iOS 14.5
- Browser: Not applicable
- Mobile device: iPhone XS

kristiyandobrev:
## Purpose

Sorting address book alphabetically. Closes #19

### DIFF
--- a/ConcordiumWallet/Views/AccountsView/Recipient/SelectRecepient/SelectRecipientPresenter.swift
+++ b/ConcordiumWallet/Views/AccountsView/Recipient/SelectRecepient/SelectRecipientPresenter.swift
@@ -106,6 +106,7 @@ class SelectRecipientPresenter {
         
         $recipients.sink { (recipients) in
             self.originalRecipientsViewModels.removeAll()
+                        
             for (index, recipient) in recipients.enumerated() {
                 let encrypted = mode == .selectRecipientFromShielded
                 self.originalRecipientsViewModels.append(RecipientViewModel(recipient: recipient, realIndex: index, isEncrypted: encrypted))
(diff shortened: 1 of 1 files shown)


ConcordiumWallet/Views/AccountsView/Recipient/SelectRecepient/SelectRecipientPresenter.swift
```diff
@@ -114,7 +115,10 @@ class SelectRecipientPresenter {
             self.originalRecipientsViewModels = Array((NSOrderedSet(array: (self.originalRecipientsViewModels)).array as? [RecipientViewModel]) ?? [])
             
             // Filter out own account.
-            self.originalRecipientsViewModels = self.originalRecipientsViewModels.filter {$0.address != ownAccount?.address}
+            self.originalRecipientsViewModels = self.originalRecipientsViewModels
+                .filter {$0.address != ownAccount?.address}
+                .sorted(by: { $0.name < $1.name })
```
abizjak, line 120: Will this ignore case or will it just order by Unicode code points?

kristiyandobrev: I believe it should be by unicode points.

abizjak: I just discussed with @jens-concordium and the intention is that this should ignore case, so presumably something like
```swift
.sorted(by: { $0.name.lowercased() < $1.name.lowercased() })
```
should be fine.

jens-concordium: Yep, as Ales mentioned the idea was to ignore the cases. I'm sorry I forgot to add it to the issue, after we discussed it.

kristiyandobrev: Sure, it is changed now.

abizjak (approved):

PR #44, opened 2021-09-13
Cleanup both identities and accounts if flow is aborted

### LINKED ISSUE #35: Aborted identity issuing flow creates identity card with unknown error
**Bug Description**
When aborting the identity issuing flow at any point after having selected the IDP and before submitting the selfie, an identity card is created.  The usual error dialog pops up, allowing the user in the newest app version to send the hashed sessionID to Notabene. In such cases, Notabene will not be able to find the sessionID. The identity card gets status "failed" with reason "An unknown error occured creating identity. Please try again". 

Android does behave more correctly. There the identity card is not created until the selfie is submitted. Therefore, no error dialog is prompted to the user neither.

Note: In both iOS and Android, the "finish" button at the end of the flow is only meant to leave the identity flow.The identity is submitted by submitting the final selfie but before clicking the "finish" button.

**Steps to Reproduce**
- select IDP
- optional: complete some steps of the identity flow
- close the flow via the X in the right upper corner
- restart the app (close and open)

**Expected Result**
- No error dialog is shown.
- No identity card is created.

**Actual Result**
- Error dialog pops up. 
- In newest draft version of app, user can send hashed sessionID to Notabene. However, Notabene won't be able to find it.
- Identity card is created as "failed" with unknown error.

**Versions**
- Software Version: 1.0.4
- OS: iOS 14.5.1
- Browser: probably Safari
- Mobile device, if applicable: iPhone 11

IMG_3621

kristiyandobrev:
## Purpose

Making sure all identities and accounts are cleaned up if the identity creation flow is aborted by the user. 

Closes #35

### DIFF
--- a/ConcordiumWallet/Views/Identities/CreateIdentity/CreateIdentityCoordinator.swift
+++ b/ConcordiumWallet/Views/Identities/CreateIdentity/CreateIdentityCoordinator.swift
@@ -104,6 +104,25 @@ class CreateIdentityCoordinator: Coordinator, ShowError {
         }
         createdIdentity = nil
     }
+    
+    private func cleanupUnfinishedIdentiesAndAccounts() {
+        cleanupUnfinishedIdenties()
+        cleanupUnfinishedAccounts()
+    }
+    
+    private func cleanupUnfinishedAccounts() {
+        let unfinishedAccounts = dependencyProvider.storageManager().getAccounts().filter { $0.address == "" || $0.transactionStatus == .committed}
+        for account in unfinishedAccounts {
+            dependencyProvider.storageManager().removeAccount(account: account)
+        }
+    }
+    
+    private func cleanupUnfinishedIdenties() {
+        let unfinishedIdentities = dependencyProvider.storageManager().getIdentities().filter { $0.ipStatusUrl.isEmpty }
+        for identity in unfinishedIdentities {
+            dependencyProvider.storageManager().removeIdentity(identity)
+        }
+    }
 
     func showInitialAccountInfo() {
         let initialAccountPresenter = InitialAccountInfoPresenter(delegate: self, type: .newAccount)
@@ -187,18 +206,11 @@ extension CreateIdentityCoordinator: CreateNicknamePresenterDelegate {
             showIdentityList(withIdentityName: nickname)
         }
     }
-    
-    private func cleanupUnfinishedAccounts() {
-        let unfinishedAccounts = dependencyProvider.storageManager().getAccounts().filter { $0.address == ""}
-        for account in unfinishedAccounts {
-            dependencyProvider.storageManager().removeAccount(account: account)
-        }
-    }
 }
 
 extension CreateIdentityCoordinator: IdentitiyProviderListPresenterDelegate {
     func closeIdentityProviderList() {
-        cleanupUnfinishedAccounts()
+        cleanupUnfinishedIdentiesAndAccounts()
         parentCoordinator?.createNewIdentityCancelled()
     }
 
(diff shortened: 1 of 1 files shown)


ConcordiumWallet/Views/Identities/CreateIdentity/CreateIdentityCoordinator.swift
```diff
@@ -104,6 +104,25 @@ class CreateIdentityCoordinator: Coordinator, ShowError {
...
         createdIdentity = nil
     }
+    
+    private func cleanupUnfinishedIdentiesAndAccounts() {
+        cleanupUnfinishedIdenties()
+        cleanupUnfinishedAccounts()
+    }
+    
+    private func cleanupUnfinishedAccounts() {
+        let unfinishedAccounts = dependencyProvider.storageManager().getAccounts().filter { $0.address == "" || $0.transactionStatus == .committed}
```
abizjak, line 114: I'm not sure I understand the logic here. Why would we want to remove accounts that have been committed (but not finalized)?

I'm also not really sure about `$0.address == ""`. This seems to be used a lot, but it is a bit unclear how. Does this check whether the account is initial?

kristiyandobrev: If the identity flow is aborted (prior adding an identity) by the user, it leaves accounts with `transactionStatus =  .commited` which leads to the error mentioned in the issue. I guess that is sort of a "temporary" state until an identity gets linked to the account.

abizjak: So what is the check `$0.address = ""` there for?

kristiyandobrev: That should check if the account is initial.

abizjak: So  with this I would expect the check to be
```
$0.address == "" && $0.transactionStatus == .committed
```
or am I missing something?

kristiyandobrev: I believe we should persist the OR sign. 

My understanding is the following. When we fist start the identity creation we create and account with empty address (i.e `address == ""`). Then as we progress we select an identity provider and the transaction status changes to commited (i.e` transactionStatus == .committed`). It is possible to abandon the flow in between and therefore we have the `OR`.  Correct me if I am wrong.

abizjak: I'm a bit concerned that this will delete accounts which should not be deleted some of the time.

In particular what will happen in the following scenario
- user has a valid identity, they go and initiate account creation from that identity
- as soon as they submit this account to the chain they go and create a new identity
- will it now delete this pending account (suppose that at this time the transaction for creating the account was committed, but not yet finalized)?

abizjak: I could be misunderstanding the code here https://github.com/Concordium/concordium-reference-wallet-ios/blob/f8d0462bb482eac21da7b33ff9a7efbee7afb99d/ConcordiumWallet/Views/AccountsView/Service/AccountsService.swift#L128

kristiyandobrev: > will it now delete this pending account (suppose that at this time the transaction for creating the account was committed, but not yet finalized)?

Correct. We risk deleting that account. I thought that the account receives the `finalized` status as soon as the identity is submitted for verification but obviously that's not the case.

Well, in that case we need to find another way to differentiate between those accounts.

kristiyandobrev: Actually `$0.address == "" && $0.transactionStatus == .committed` would be enough.

abizjak (approved):

PR #45, opened 2021-09-13
Handle camera permissions during identity creation flow

### LINKED ISSUE #24: Black camera view after denying permission
**Bug Description**
A completely black camera view is shown when you deny the request to use the camera.
The app should check whether the camera is enabled and never show the black camera view.
NB: Users can manually enable the camera permissions in Settings > Concordium

**Steps to Reproduce**
1. Install the wallet
2. When creating an identity, the app will ask for permission to use the camera: click 'Don't allow'
3. Try to take a photo of your id document

**Expected Result**
The app will tell you that it is not possible to create an identity without the use of a camera and prompt you for permissions again.

**Actual Result**
The camera opens up, but everything is black:

**Versions**
- Software Version: 10.0.14 (testnet), 1.0.4 (mainnet)
- OS: 14.6
- Browser: Safari / Brave Browser (not sure which one is used when a webpage is loaded within an app)
- Mobile device, if applicable: iPhone 11 Pro

kristiyandobrev:
## Purpose

Prompting error dialog when choosing identity provider (during identity creation flow) if the camera permissions are either `denied` or `restricted`

Closes #24


ConcordiumWallet/Resources/ConcordiumWallet/Base.lproj/Localizable.strings
```diff
@@ -34,6 +34,7 @@ add_identity_data = "Reveal Identity Data";
 "viewError.duplicateRecipient" = "This account was not added to the address book, as it was already saved as ";
 "viewError.passwordChangeFailed" = "An error occured while changing password";
 "viewError.jailbroken" = "This iPhone might be jailbroken. Using this iPhone is not safe.";
+"view.error.cameraAccessDenued" = "Camera access has been denied, and must be enabled in the Settings app to continue";
```
abizjak, line 37: Should this be `cameraAccessDenied` instead?

ConcordiumWallet/Service/ErrorMapper.swift
```diff
@@ -45,6 +46,8 @@ extension ViewError: LocalizedError {
             return "viewError.duplicateRecipient".localized + name
         case .exportUnfinalizedAccounts(let unfinalizedAccountsNames):
             return "export.unfinalizedAccounts".localized + unfinalizedAccountsNames.joined(separator: ", ")
+        case .cameraAccessDeniedError:
+            return "view.error.cameraAccessDenued".localized
```
abizjak, line 50: And correspondingly here as well.

ConcordiumWallet/Views/Identities/CreateIdentity/IdentityProviderList/IdentityProviderListPresenter.swift
```diff
@@ -121,6 +122,11 @@ extension IdentityProviderListPresenter: IdentityProviderListPresenterProtocol {
     }
     
     func userSelected(identityProviderIndex: Int) {
+        guard CameraHelper.isCameraAccessAllowed() else {
+            view?.showRecoverableAlert(.cameraAccessDeniedError) { SettingsHelper.openAppSettings() }
```
abizjak, line 126: Could the app here also ask for the permission (as opposed to just checking it)?

With this check, if the user gets to the identity verification flow and then denies camera permission, will they just end up with a blank screen as they do now?

The onfido flow does not handle denying the permission very well, so ideally the app would acquire the permission before entering the identity issuer flow.

ConcordiumWallet/Resources/ConcordiumWallet/Base.lproj/Localizable.strings
kristiyandobrev: It should 😁

ConcordiumWallet/Views/Identities/CreateIdentity/IdentityProviderList/IdentityProviderListPresenter.swift
kristiyandobrev: Yep, we could do that as well. I will push the necessary changes :)

abizjak (approved):

PR #47, opened 2021-09-15
Storing commitments randomness

### LINKED ISSUE #23: Store account randomness securely in database
**Task description**
In the future, users should be able to reveal more attributes on their accounts at a later time than on account creation. To enable this, the randomness used to reveal attributes on account creation must be saved in the database so it can be accessed later. 

**Requirements:**

- The mobile wallet saves the randomness for new accounts in the database in a secure way.
- The randomness can be accessed later, so people can reveal more attributes on their account. We should also have the option to access the randomness for potential future off-chain use-cases.
- The new version of the mobile wallet must be able to handle “old” accounts created before the randomness could be saved in the database. 
- The export/import feature must support the saved randomness.

**Notes:**

- The randomness is secret material so it must be handled in a secure way, on par with account keys  and IdCredSec.
- The response from createCredential from the library supplied by Concordium, has an additional field commitmentsRandomness.

kristiyandobrev:
## Purpose

Storing `commitmentsRandomness` object securely. 

Key takeways: 

- Enabled database schema versioning and migrations
- Introduced `RealmHelper` for handling database migrations
- Storing `commitmentsRandomness` object in the same manner as account keys and encryption secret
- Auto-generated `CommitmentsRandomness` models via **QuickType**

Closes #23


abizjak (approved):
I can't really comment on the iOS parts and the UX flows, but the modelling of the randomness looks right.

PR #57, opened 2021-09-27
Transaction memos

### LINKED ISSUE #34: New transaction types with memos
**Task description**
The sender should be able to attach text to transfers that will enable him to convey additional information to the receiver. The current use-case mandating this is ease of integration with exchanges.

**Sub-Tasks**
* Support sending transactions with memos
* Support displaying memos in the transaction list; existing wallets will not display memos on incoming transfers, but will otherwise continue to function fully as they do now.
* Support for new wallet-poxy

**New Transfer Types**
* simple account-to-account transfers
* scheduled transfers
* encrypted transfers

**Backwards Compatibility**
The change must be backwards compatible. Existing transactions must continue to be accepted by the new protocol, and they must have the same semantics to the extent possible.

**EDIT:** The flow specifying how this should work, can be found in Figma in "Flow 19"

kristiyandobrev:
## Purpose

Closes #34


ConcordiumWallet/Model/Memo.swift
```diff
@@ -0,0 +1,59 @@
...
+        guard
+            let hex = hex,
+            let data = Data(hex: hex)
+        else {
+            return nil
+        }
+        
+        let bytes = data.bytes
+        guard case .utf8String(let value) = try? CBOR.decode(bytes) else { return nil }
+        self.displayValue = value
```
abizjak, line 52: I'm not that well-versed in Swift, what will this do if the data cannot be decoded as string? Will it just be null?

I think the idea was to display hex as is in case it cannot be decoded as a valid string.

Additionally, does `decode` make sure that the entire `bytes` array is consumed, e.g., suppose that the prefix of the byte array is valid CBOR, but there is additional data. Will `CBOR.decode` succeed?

kristiyandobrev: It will produce a nil and therefore would not be displayed (as it is right now). 

> I think the idea was to display hex as is in case it cannot be decoded as a valid string.

Sure. I can change it easily. 

> Will CBOR.decode succeed?

It will.

abizjak: Regarding partial decoding, I think that is wrong. It should either be valid CBOR or not, the decoding/display should not omit information due a partial parse. Is there a different `decode` method or a way to check that the entire input was decoded?

kristiyandobrev: Ah, sorry. I misinterpret it. I meant invalid CBOR will fail to decode.

abizjak: Ah, OK.

abizjak (approved):

PR #60, opened 2021-09-27
Accounts identity cleanup

### LINKED ISSUE #24: Black camera view after denying permission

kristiyandobrev:
## Purpose

* Collect all cleanup funcs in `StorageManager`
* Cleanup accounts without address upon login*
* Bump project version (18)

_* That will cover the scenario when the user changes a permission from the `Settings` app and the iOS restarts the app. Unfortunately iOS does not provide us with such a system event and the best we can do is to basically cleanup the accounts without address upon login_

Closes #24

### DIFF
--- a/ConcordiumWallet/Service/StorageManager.swift
+++ b/ConcordiumWallet/Service/StorageManager.swift
@@ -52,6 +52,11 @@ protocol StorageManagerProtocol {
     func getLastEncryptedBalanceTransfer(for accountAddress: String) -> TransferDataType?
     func getAllTransfers() -> [TransferDataType]
     func removeTransfer(_ transfer: TransferDataType?)
+    
+    func removeUnfinishedIdentities()
+    func removeUnfinishedAccounts()
+    func removeAccountsWithoutAddress()
+    func removeUnfinishedAccountsAndRelatedIdentities()
 }
 
 enum StorageError: Error {
@@ -61,7 +66,6 @@ enum StorageError: Error {
 }
 
 class StorageManager: StorageManagerProtocol {
-
     private var realm: Realm = try! Realm(configuration: RealmHelper.realmConfiguration) // swiftlint:disable:this force_try
     private var keychain: KeychainWrapperProtocol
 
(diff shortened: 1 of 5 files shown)


ConcordiumWallet/Service/StorageManager.swift
```diff
@@ -373,4 +377,38 @@ class StorageManager: StorageManagerProtocol {
...
+    func removeUnfinishedIdentities() {
+        let unfinishedIdentities = getIdentities().filter { $0.ipStatusUrl.isEmpty }
+        for identity in unfinishedIdentities {
+            removeIdentity(identity)
+        }
+    }
+    
+    func removeUnfinishedAccounts() {
+        guard
+            let unfinishedAccount = getAccounts().first(where: { $0.identity == nil || $0.identity?.ipStatusUrl == nil || $0.identity?.state == .failed })
```
abizjak, line 390: Why only remove the first account? The function name indicates it should remove all unfinished accounts.

ConcordiumWallet/Service/StorageManager.swift
```diff
@@ -373,4 +377,38 @@ class StorageManager: StorageManagerProtocol {
             realm.delete(transferEntity)
         }
     }
+    
+    func removeUnfinishedIdentities() {
+        let unfinishedIdentities = getIdentities().filter { $0.ipStatusUrl.isEmpty }
```
abizjak, line 382: Are failed identities removed elsewhere?

ConcordiumWallet/Service/StorageManager.swift
kristiyandobrev: There has been no change in that since our last discussion in [here](https://github.com/Concordium/concordium-reference-wallet-ios/pull/44). In this PR the functions for cleaning up accounts and identities were only moved to another more centeralized place - `StorageManager`

abizjak: Ok.

abizjak (approved):
Ok, since this is just refactoring.

I am still not sure about the entire handling, but this PR does not seem to introduce any observable change.

kristiyandobrev:
> Ok, since this is just refactoring.
> 
> I am still not sure about the entire handling, but this PR does not seem to introduce any observable change.

It does in [here](ConcordiumWallet/AppCoordinator.swift) actually. As the description above says "Cleanup accounts without address upon login"

Issue #37, opened 2021-09-07
Initial account ends up "failed" in iOS, but it is actually finalised on chain


jens-concordium:
**Bug Description**

In some rare cases an initial account is perceived as failed by the iOS app, while the initial account has actually been finalised on the chain.

We have pieced together one scenario, that seemed to go like this, but some parts are assumptions:

1. ~ 15:34 User submits identity issuance request to Notabene. The iOS mobile wallet queries the wallet proxy for the account immediately after submitting the request to Notabene, and long before the account has actually been finalised or even submitted to the chain. (EDIT: As mentioned in a below comment, the fact that the mobile wallet queries the account at this stage is not according to the agreed protocol, and should ideally be fixed as well.)
2. Shortly after 15:34 (Assumed, not certain): Notabene submits request to Onfido to get the ID verified.
3. Somewhere between 15:34 and 16:32:19 Onfido verifies the ID and then sometime just before 16:32:19 Notabene submits the initial account to the wallet proxy.
4. The user / mobile wallet queries the account balance at 16:32:13, the wallet does not exist yet, as the account was not created before 16:32:19. (Regarding the timing: Maybe the user wanted to wait an hour before checking if it succeeded? )
5. The user / mobile wallet queries the account again at 16:32:20, it does now exist (1 sec after the creation block by pure coincidence?) but it is not finalised yet.
6. At 16:32:28 Notabene queries the initial account transaction hash to see if it finalised. It is not entirely certain what the response from the wallet proxy was, but based on the size it could be either "committed" or "finalised". It did take two blocks before the initial account was actually finalised though, which could indicate that the response Notabene got was "committed".
7. The account balance was never queried again by the mobile wallet, so there is reason to expect that it had perceived the initial account as failed, hence removing it from the app.

This scenario might hint that Notabene's service knew that the initial account status was "committed", and returned this status to the mobile wallet, which did not know how to interpret this, and deemed the account failed. We cannot know for sure though, so this should be seen as a suggestion of what might cause the problem.

We need to investigate if this is actually the case, or if something else is causing the perceived failed initial accounts in iOS. 

**Steps to Reproduce**
There does not seem to be any certain way to reproduce this.

**Expected Result**
The initial account should not be regarded as failed in the mobile wallet, before it is absolutely certain that it has failed.

**Actual Result**
The initial account is regarded as failed by the mobile wallet, while it shouldn't be.

**Versions**
[...]

Issue #33, opened 2021-08-10
Wallet accounts missing on chain (iOS repo)


concordium-cl:
**Bug Description**

We have received several support tickets where users have created identities and initial accounts successfully, i.e. identity and account cards look correct with green checkmark, but the account is not on the chain.

Examples:
* 3eMpGKfajwgzjUbfsVFou4ZoR3ALNYbe5w5VmNJ3LY5UoNXQsw
* 4ckJBDGyxMtRbXL3BAKscZkMDjxkXDLF4wKTaZjy29R2g87KEr
* 4miVxofGUV4AeXM3zauGwkbDxHUtvKhoxa1EpgicU7keZggj69

**Steps to Reproduce**
Unclear.

**Expected Result**
Either accounts fail or they can be found on chain.

**Actual Result**
Successful accounts cannot be found on chain.

**Versions**
- iOS 1.0.4

concordium-cl:
See also https://github.com/Concordium/notabene/issues/14

jens-concordium:
@concordium-cl 

There is also no real way to test and verify this, as we cannot reproduce it. Thomas' fix may also have helped here, but I think we also have to keep an eye out for new reports on this, to see if it has helped.

PR #61, opened 2021-09-28
Refactor identity creation process.

### LINKED ISSUE #53: Letting the app lock the user out during identity creation, creates identity card with unknown error.
**Bug Description**
Much like #35, letting the app lock out the user during the identity creation flow, will create a failed identity card with unknown error. The identity card will show the hashed session ID, which technically doesn't exist as it has not been submitted to Notabene. 

Waiting for long enough to let the app trigger FaceID / passcode prompt, should not create a failed identity card.

**Steps to Reproduce**
1. Start making a new identity, and browse long enough to get to the identity provider selection screen.
2. Let the app wait long enough for FaceID / passcode prompt to trigger.
3. Notice how pressing Cancel will yield a failed identity card which should not be there.

**Expected Result**
After the app asks for FaceID / passcode, the user should just continue from the current screen in the flow. No failed identity should be created.

**Actual Result**
After the FaceID / passcode prompt, the user is taken out of the flow and a failed identity exists, which shouldn't be there. 

**Versions**
- Software Version: 1.0(16)
- OS: iOS 14.7.1
- Mobile device: iPhone XS

td202:
## Purpose

This PR refactors identity creation so that identity and initial account objects are not created until the identity issuance flow is completed. Prior to this, temporary account objects are created as part of the identity issuance process, and account and identity objects are created before the issuance flow completes, which may cause them to show in the account list, even when they cannot be created.

This should give a more robust solution to #35 and should fix #53. It is possible this could help with #33, since the previous process could have led to confusion between accounts in some circumstances where the identity flow is interrupted and restarted. It may also help with #37, although more investigation is required there.

## Changes

- Introduce an `IdentityCreation` class that stores the details used for identity creation temporarily before and during the interaction with the identity provider. This is not stored in the realm database.
- Pass around the name used for the initial account prior to its creation rather than storing it with a temporary account in the realm.
- Create an `IdentityCreation` immediately before invoking the identity provider, rather then creating an account and identity object at this point. This ensures that the account and identity aren't created prematurely. A UUID identifying the `IdentityCreationEntity` is added to the callback URL to ensure that there cannot be confusion about which identity is verified.
- Defer creating the initial account and identity until after the callback from the identity provider occurs, using the UUID to ensure that the correct account and identity objects are created.

## Checklist

      hard-to-understand areas.


ConcordiumWallet/Service/MobileWallet.swift
```diff
@@ -44,6 +46,80 @@ enum MobileWalletError: Error {
     case invalidArgument
 }
 
+final class IdentityCreation {
```
kristiyandobrev, line 49: I think it would be a good idea if that class gets a file of it's own.

ConcordiumWallet/Service/MobileWallet.swift
```diff
@@ -44,6 +46,80 @@ enum MobileWalletError: Error {
...
+    /// Identifier used in the callback to identify this identity creation
+    let id = NSUUID().uuidString
```
kristiyandobrev, line 51: You might replace `NSUUID` (Objc implementation) with `UUID` (Swift implementation)

ConcordiumWallet/Service/MobileWallet.swift
```diff
@@ -102,46 +174,15 @@ class MobileWallet: MobileWalletProtocol {
...
-                        _ = identity.withUpdated(encryptedPrivateIdObjectData: $0)
-                        let newAccount = account.withUpdatedIdentity(identity: identity)
-                        let shieldedAmount = ShieldedAmountTypeFactory.create().withInitialValue(for: newAccount)
-                        _ = try? self.storageManager.storeShieldedAmount(amount: shieldedAmount)
-                    }.publisher.eraseToAnyPublisher()
-            }
-            .map { _ in
-                data.idObjectRequest
-            }.eraseToAnyPublisher()
+                .tryMap { (pwHash) in
```
kristiyandobrev, line 177: Since you are in a closure you and you reference `self` down below you need to capture it weakly: 

```suggestion
                .tryMap { [weak self] (pwHash) in
```

ConcordiumWallet/Service/MobileWallet.swift
```diff
@@ -102,46 +174,15 @@ class MobileWallet: MobileWalletProtocol {
...
+                    let identityCreation = try IdentityCreation(initialAccountName: initialAccountName,
+                                                            identityName: identityName,
+                                                            identityProvider: identityProvider,
+                                                            data: data,
+                                                            pwHash: pwHash,
+                                                            storageManager: self.storageManager)
```
kristiyandobrev, line 183: Since you capture the `self` weakly it becomes optional:

```suggestion
                                                            storageManager: self?.storageManager)
```

ConcordiumWallet/Views/Identities/CreateIdentity/CreateIdentityCoordinator.swift
```diff
@@ -67,24 +76,50 @@ class CreateIdentityCoordinator: Coordinator, ShowAlert {
...
+        newAccount.encryptedAccountData = createdIdentity.encryptedAccountData
+        newAccount.encryptedPrivateKey = createdIdentity.encryptedPrivateKey
+        newAccount.identity = newIdentity
+        _ = try storageManager.storeAccount(newAccount)
+        let shieldedAmount = ShieldedAmountTypeFactory.create().withInitialValue(for: newAccount)
+        _ = try storageManager.storeShieldedAmount(amount: shieldedAmount)
+        
+        self.createdIdentity = nil
+        
+        navigationController.dismiss(animated: true) {
```
kristiyandobrev, line 120:
```suggestion
        navigationController.dismiss(animated: true) { [weak self] in
```

ConcordiumWallet/Views/Identities/CreateIdentity/CreateIdentityCoordinator.swift
```diff
@@ -67,24 +76,50 @@ class CreateIdentityCoordinator: Coordinator, ShowAlert {
...
+            self.showIdentitySubmitted(identity: newIdentity, account: newAccount)
```
kristiyandobrev, line 121:
```suggestion
            self?.showIdentitySubmitted(identity: newIdentity, account: newAccount)
```

ConcordiumWallet/Views/Identities/CreateIdentity/CreateIdentityCoordinator.swift
```diff
@@ -109,14 +144,15 @@ class CreateIdentityCoordinator: Coordinator, ShowAlert {
             parentCoordinator?.createNewIdentityCancelled()
         } else {
             let serverError = ViewError.simpleError(localizedReason: error.error.detail)
-            createIdentityFailed(serverError)
+            navigationController.dismiss(animated: true) {
```
kristiyandobrev, line 147:
```suggestion
            navigationController.dismiss(animated: true) { [weak self] in
```

ConcordiumWallet/Views/Identities/CreateIdentity/CreateIdentityCoordinator.swift
```diff
@@ -109,14 +144,15 @@ class CreateIdentityCoordinator: Coordinator, ShowAlert {
...
+                self.showFailedIdentityCreation(error: serverError)
```
kristiyandobrev, line 148:
```suggestion
                self?.showFailedIdentityCreation(error: serverError)
```

ConcordiumWallet/Views/Identities/CreateIdentity/IdentityProviderList/IdentityProviderListPresenter.swift
```diff
@@ -141,27 +143,33 @@ extension IdentityProviderListPresenter: IdentityProviderListPresenterProtocol {
...
             }
             
-            self.identity = self.identity.withUpdated(identityProvider: IdentityProviderDataTypeFactory.create(ipData: ipInfoResponse))
+            let identityProvider = IdentityProviderDataTypeFactory.create(ipData: ipInfoResponse)
             
             let wallet = self.dependencyProvider.mobileWallet()
             
             self.service.getGlobal().flatMap { global in
-                wallet.createIdRequestAndPrivateData(identity: self.identity, global: global, requestPasswordDelegate: delegate)
+                wallet.createIdRequestAndPrivateData(initialAccountName: self.initialAccountName,
```
kristiyandobrev, line 151: Weak self is needed above 👆

kristiyandobrev (changes requested):
I would also suggest rebasing onto the latest `main` branch.

ConcordiumWallet/Service/MobileWallet.swift
td202: I'm not convinced that this is necessary. (`self` was used previously without being captured weakly.) My understanding is that this is to avoid a retain cycle, leading to a memory leak. I think this cannot happen or doesn't matter for a couple of reasons:
1. `MobileWallet` does not retain the publisher itself, and the caller calls `sink` on the result, so eventually the closure should no longer be retained.
2. The `MobileWallet` instance is retained for the entire lifetime of the application. This means that a retain cycle involving it is irrelevant, since anything that is retained indefinitely by `MobileWallet` will be retained indefinitely whether or not it also retains `MobileWallet`.

On the other hand, `weak self` would mean that we would have to account for the (impossible) case where `self` becomes `nil`.

kristiyandobrev: Ah, yeah. I forgot that `MobileWallet` is injected by the `ServicesProvider`. My bad, sorry 🙋🏻‍♂️

kristiyandobrev (approved):

PR #74, opened 2021-10-12
Handle partially valid memos

### LINKED ISSUE #66: Information in memo-raw files is wrongly interpreted
**_Related to task https://github.com/Concordium/concordium-reference-wallet-ios/issues/34_**

**Bug Description**
For a memo.txt with content _this is a memo text in a file_ some characters are wrongly interpreted as type and length.

Below is the file and the old network dashboard output.
memo file

The iOS output is exactly the same:
IMG_4013

@abizjak explanation: _The letter t's ASCII code is 116 which is 01110100 in binary. The way CBOR encoding works is it looks at the first 3 bits to determine the type, in this case the type is determined to be (3=011 ) which is string. The remaining 5 bits are 10100 which is 20 in decimal, which is interpreted as the lenght of the remaining string. Hence the remaining 20 bytes are decoded._

**Steps to Reproduce**
- do a scheduled transfer with `--memo-raw memo.txt`

**Expected Result**
The memo should be printed as hex value in case of invalid CBOR.

**Actual Result**
See above.

**Versions**
- 1.0.18
- iOS 14.7
- iPhone 11

kristiyandobrev:
## Purpose

Screenshot 2021-10-12 at 13 59 11

Handling partially valid memos (mentioned [here](#66)) by showing hex value instead of what appears to be partially decoded value.

Closes #66

## Changes

Re-encoding of hex value and then comparing the length to the actual hex. If the two values are equal we have a success if not we assign the original `hex` string as `displayValue`. 

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/ConcordiumWallet/Model/Memo.swift
+++ b/ConcordiumWallet/Model/Memo.swift
@@ -42,14 +42,19 @@ struct Memo: MemoDataType {
     init?(hex: String?) {
         guard
             let hex = hex,
-            let data = Data(hex: hex)
+            let data = Data(hex: hex),
+            case .utf8String(let decodedValue) = try? CBOR.decode(data.bytes)
         else {
             return nil
         }
         
-        let bytes = data.bytes
-        guard case .utf8String(let value) = try? CBOR.decode(bytes) else { return nil }
-        self.displayValue = value
+        let decodedValueHex = Data(bytes: decodedValue.encode(), count: decodedValue.encode().count).hexDescription
+        
+        if decodedValueHex == hex {
+            self.displayValue = decodedValue
+        } else {
+            self.displayValue = hex
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


ConcordiumWallet/Model/Memo.swift
```diff
@@ -42,14 +42,19 @@ struct Memo: MemoDataType {
...
+            let data = Data(hex: hex),
+            case .utf8String(let decodedValue) = try? CBOR.decode(data.bytes)
         else {
             return nil
         }
         
-        let bytes = data.bytes
-        guard case .utf8String(let value) = try? CBOR.decode(bytes) else { return nil }
-        self.displayValue = value
+        let decodedValueHex = Data(bytes: decodedValue.encode(), count: decodedValue.encode().count).hexDescription
```
abizjak, line 51: Perhaps it would be better to use `.encode` only once to save effort.

I am also wondering why the constructor needs the length at all?

ConcordiumWallet/Model/Memo.swift
```diff
@@ -42,14 +42,19 @@ struct Memo: MemoDataType {
...
+        
+        if decodedValueHex == hex {
+            self.displayValue = decodedValue
+        } else {
+            self.displayValue = hex
+        }
```
abizjak, line 57: I would urge you to document why this relatively silly thing is being done (as we discussed in slack).

PR #82, opened 2021-10-29
Notification on succesful new account 

### LINKED ISSUE #79: Notification on succesful new account
**Description**
Currently when an account is finalized, the user is not given any explicit notification. People have been asking for more explicit notifications, and while we cannot push notifications to the OS at the moment, we could show one inside the app after polling the wallet proxy.

This notification could also be used to suggest that people should make a backup after a new account has been added, just to encourage backups more often. 

So the suggested change would be:
- Show a pop-up when an account has finalized, that also reminds the user to make a backup. The backup then has two buttons that say "Okay" and "Make backup". Okay will just close the pop-up and leave the user on the current page, and Make backup would take them to the export flow.

Sketches showing the change, can be found in Figma.

**Versions**
- Software Version: 1.0(20)
- OS: iOS

kristiyandobrev:
## Purpose

Screenshot 2021-10-29 at 13 52 02

Closes #79 

## Changes

- Saving pending accounts addresses in user preferences (that will allow showing the notification across app sessions)
- Checking regularly (on view did appear event) if the pending accounts are being finalized
- Show notification alert when an account is being finalized

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/ConcordiumWallet.xcodeproj/project.pbxproj
+++ b/ConcordiumWallet.xcodeproj/project.pbxproj
@@ -4954,7 +4954,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = "ConcordiumWallet/Resources/Entitlements/Concordium ID.entitlements";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 21;
+				CURRENT_PROJECT_VERSION = 22;
 				DEVELOPMENT_TEAM = K762RM4LQ3;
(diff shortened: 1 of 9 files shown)


kristiyandobrev:
> Looks good and seems to work fine. I'm a little concerned about the code duplication between AccountsCoordinator and MoreCoordinator from a maintenance perspective, but it's fine.

I agree about the duplication. Is not the best solution but in fact this is how the coordinator pattern forces us to do things and in my opinion it's main drawback. I would much rather prefer something like "Router" pattern which seems more of a cleaner approach 😊

PR #98, opened 2021-11-11
Finalized account alerts fixes 

kristiyandobrev:
## Purpose

* Delivering batch notification when multiple accounts are finalized https://github.com/Concordium/concordium-reference-wallet-ios/issues/79#issuecomment-964864489

* Fix issue with the passcode's navbar color if the app locks during the identity/account creation process. 

|Screenshot 2021-11-11 at 09 06 14|
|-|

## Changes

* Zipping all network requests for accounts statuses instead of firing them in a for loop
* `UIViewController` extension var for checking whether the controller is shown on screen or not. Particularly useful to decide whether or not to show finalized account notification 
* `AccountSubmissionStatus` wrapper model
* Set light preferred status bar style for pass code view (previously dark) 
* Apply transparent navbar appearance upon entering passcode view (otherwise it pick the parent navigation controller and it might become white like the screenshot above)
* 
## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/ConcordiumWallet/Resources/ConcordiumWallet/Base.lproj/Localizable.strings
+++ b/ConcordiumWallet/Resources/ConcordiumWallet/Base.lproj/Localizable.strings
@@ -400,7 +400,10 @@ idiss@concordium.software";
 
 "copyreference.info.text" = "You can copy the reference below, and then contact Notabene at concordium-idiss@notabene.id \n\n Optional: Add Concordium as CC via idiss@concordium.software";
 
-"accountfinalized.alert.title" = "Your new account is ready!";
-"accountfinalized.alert.message" = "Your new account %@ has finalized, and is now ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
+
+"accountfinalized.single.alert.title" = "Your new account is ready!";
+"accountfinalized.single.alert.message" = "Your new account %@ has finalized, and is now ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
+"accountfinalized.multiple.alert.title" = "Your new accounts are ready!";
+"accountfinalized.multiple.alert.message" = "Your new accounts are now finalized, and ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
 "accountfinalized.alert.action.backup" = "Make back-up";
 
--- a/ConcordiumWallet/Resources/ConcordiumWallet/Base.lproj/Localizable.strings
+++ b/ConcordiumWallet/Resources/ConcordiumWallet/Base.lproj/Localizable.strings
@@ -402,7 +402,7 @@ idiss@concordium.software";
 
 
 "accountfinalized.single.alert.title" = "Your new account is ready!";
-"accountfinalized.single.alert.message" = "Your new account %@ has finalized, and is now ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
+"accountfinalized.single.alert.message" = "Your new account %@ has been finalized, and is now ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
 "accountfinalized.multiple.alert.title" = "Your new accounts are ready!";
 "accountfinalized.multiple.alert.message" = "Your new accounts are now finalized, and ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
 "accountfinalized.alert.action.backup" = "Make back-up";
--- a/ConcordiumWallet/Service/StorageManager.swift
+++ b/ConcordiumWallet/Service/StorageManager.swift
@@ -452,11 +452,10 @@ class StorageManager: StorageManagerProtocol { // swiftlint:disable:this type_bo
     }
     
     func storePendingAccount(with address: String) {
-        guard !getPendingAccountsAddresses().contains(where: { $0 == address }) else { return }
-
         let key = UserDefaultKeys.pendingAccount.rawValue
-        
+
         if var pendingAccounts = UserDefaults.standard.stringArray(forKey: key) {
+            guard !pendingAccounts.contains(where: { $0 == address }) else { return }
             pendingAccounts.append(address)
             UserDefaults.standard.set(pendingAccounts, forKey: key)
         } else {
--- a/ConcordiumWallet/Service/StorageManager.swift
+++ b/ConcordiumWallet/Service/StorageManager.swift
@@ -452,6 +452,8 @@ class StorageManager: StorageManagerProtocol { // swiftlint:disable:this type_bo
     }
     
     func storePendingAccount(with address: String) {
+        guard !getPendingAccountsAddresses().contains(where: { $0 == address }) else { return }
+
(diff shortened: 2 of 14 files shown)


td202 (commented):
If I import a back-up without having created any accounts, it shows the "Your new accounts are ready" message, which I don't think should be the case.

ConcordiumWallet/Resources/ConcordiumWallet/Base.lproj/Localizable.strings
```diff
@@ -400,7 +400,10 @@ idiss@concordium.software";
 
 "copyreference.info.text" = "You can copy the reference below, and then contact Notabene at concordium-idiss@notabene.id \n\n Optional: Add Concordium as CC via idiss@concordium.software";
 
-"accountfinalized.alert.title" = "Your new account is ready!";
-"accountfinalized.alert.message" = "Your new account %@ has finalized, and is now ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
+
+"accountfinalized.single.alert.title" = "Your new account is ready!";
+"accountfinalized.single.alert.message" = "Your new account %@ has finalized, and is now ready for use.\n\nWe recommend keeping a back-up of all accounts. You can make one now below, or later via the More page.\n\nRemember to store your back-up securely.";
```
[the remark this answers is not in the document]
kristiyandobrev: @td202 I have copied the text from the Figma file. Please confirm the change with @jens-concordium.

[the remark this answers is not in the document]
jens-concordium: Fine by me!

kristiyandobrev:
@td202 

I am unable to reproduce that actually 🤔

Here is what I did: 

1. Create account and wait until it gets verified. 
2. Export the account.
3. Uninstall the app. 
4. Install the app and start the account creation flow. 
5. Import the exported account from step 2 instead of creating new account.
6. Account is imported successfully.
7. No notification is shown.  

I know for a fact that non-finalized accounts cannot be exported. For some reason the accounts in your import has got marked as "pending" by [this](https://github.com/Concordium/concordium-reference-wallet-ios/blob/main/ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift#L270-L276) function. This means that after they got imported they got a status of either `committed` or `received` which seems strange to me. I am not entirely sure how that happens. @abizjak do you have any idea ?

concordium-cl:
@kristiyandobrev: During testing last week (on version 1.0.22), what @td202 describes above happened consistently to me whenever I imported a backup. The notification mentioned a random account "to be ready now", even if that account had GTU on it already (and thereby was finalized for sure).

concordium-cl:
@abizjak, @kristiyandobrev needs your help in that, see his comment above.

td202:
> I know for a fact that non-finalized accounts cannot be exported. For some reason the accounts in your import has got marked as "pending" by [this](https://github.com/Concordium/concordium-reference-wallet-ios/blob/main/ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift#L270-L276) function. This means that after they got imported they got a status of either `committed` or `received` which seems strange to me. I am not entirely sure how that happens. @abizjak do you have any idea ?

It looks like this could be responsible:

https://github.com/Concordium/concordium-reference-wallet-ios/blob/c9edf0822f9dabe19ecc15b3a07dda665b74773f/ConcordiumWallet/Views/MoreSection/Import/IdentityImporter.swift#L81

td202:
> I know for a fact that non-finalized accounts cannot be exported. For some reason the accounts in your import has got marked as "pending" by [this](https://github.com/Concordium/concordium-reference-wallet-ios/blob/main/ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift#L270-L276) function. This means that after they got imported they got a status of either `committed` or `received` which seems strange to me. I am not entirely sure how that happens. @abizjak do you have any idea ?

It may perhaps only be an issue for non-initial accounts. It looks like imported accounts are initially set with the transactionStatus `.receive`, but initial accounts are later set to `.finalized`.

https://github.com/Concordium/concordium-reference-wallet-ios/blob/c9edf0822f9dabe19ecc15b3a07dda665b74773f/ConcordiumWallet/Views/MoreSection/Import/IdentityImporter.swift#L81

kristiyandobrev:
@td202 

Very good point ! Is there any reason for them being set differently ? If not I think we should align them.

td202:
@kristiyandobrev I don't see any reason for the difference, since only finalized accounts are exported.

ConcordiumWallet/Views/AccountsView/AccountsViewController.swift
```diff
@@ -151,7 +156,14 @@ class AccountsViewController: BaseViewController, Storyboarded, AccountsViewProt
...
         }.store(in: &cancellables)
-        
+
+        viewModel.$finalizedAccountsNotificationState
+            .receive(on: DispatchQueue.main)
+            .sink(receiveValue: { [weak self] state in
+                self?.showAccountFinalizedNotificationIfNeeded(state)
+            })
+            .store(in: &cancellables)
+
```
td202, line 166: This approach to handling the accounts becoming finalized seems needlessly convoluted. A field in the view model is updated to indicate that an account has been finalized. Updates to this field are monitored here, and the closure shows the finalized notification. However, the state is not really needed as part of the model; it's just needed for showing the message and can be discarded afterwards.

Wouldn't it be better then instead of using the `viewModel` as an indirect means to just call `showAccountFinalizedNotificationIfNeeded` from `updateFinalizedAccountsViewState`? (Of course, these should probably be renamed appropriately in case of such a change.) This could probably be done by adding a function to the `AccountsViewProtocol` that is implemented in `AccountsViewController` by invoking `showAccountFinalizedNotificationIfNeeded` appropriately (I guess using `DispatchQueue.main` to schedule it?)

kristiyandobrev: Makes sense. Changes applied.

ConcordiumWallet/Views/Login/RequestPasswordDelegate.swift
```diff
@@ -16,19 +16,19 @@ extension RequestPasswordDelegate where Self: Coordinator {
         var modalPasswordVCShown = false
 
         let enterPasswordController = EnterPasswordFactory.create(with: requestPasswordPresenter)
-        let enterPasswordTransperantController = TransparentNavigationController()
-        enterPasswordTransperantController.modalPresentationStyle = .fullScreen
-        enterPasswordTransperantController.viewControllers = [enterPasswordController]
+        let enterPasswordTransparantController = TransparentNavigationController()
```
td202, line 19: Transparant -> Transparent

ConcordiumWallet/Views/AccountsView/AccountsPresenter.swift
```diff
@@ -112,6 +112,11 @@ enum AccountsUIState {
     case showAccounts
 }
 
+enum FinalizedAccountsNotificationState {
```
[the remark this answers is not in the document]
kristiyandobrev: I would keep the enum as there is some more explicitness and clarity about what is happening compared to the optional string. Renaming of the enum is fine :)

ConcordiumWallet/Service/StorageManager.swift
```diff
@@ -452,6 +452,8 @@ class StorageManager: StorageManagerProtocol { // swiftlint:disable:this type_bo
     }
     
     func storePendingAccount(with address: String) {
+        guard !getPendingAccountsAddresses().contains(where: { $0 == address }) else { return }
```
td202, line 455: Suggest moving this inside the `if` below and using `pendingAccounts` instead of `getPendingAccountsAddresses`, since we're already looking it up.

td202 (approved):

Issue #104, opened 2021-11-12
Line under account name input field and Continue button can disappear


jens-concordium:
**Bug Description**
When making a new account and entering a name for it, the line under the input field will disappear below the keyboard, if you go forward and then back again. It might be good to put both the button and line above the keyboard. This is likely related to the removed dependency during the update to the iOS 15 SDK. 

![image](https://user-images.githubusercontent.com/76045015/141476756-3ccf40dd-51f6-4ff4-be87-dd3db5c9a548.png)

**Steps to Reproduce**
- Start making a new account
- Enter a name for the account
- Go forward one page, and then go back again. 
- Notice how the line under the input field has appeared, and the Continue button is also nowhere to be seen.

**Expected Result**
- Both the line and the button stays above the keyboard.

**Actual Result**
- The line and the button seems to be hidden behind the keyboard.

**Versions**
- Software Version: 1.0(22)
- OS: iOS 14.7 and 15

jens-concordium:
@kristiyandobrev, it should be possible to hide the keyboard by click somewhere outside the input field now, right? I can't get it to work in all places, on iOS 15.0. The input field kind of slides down behind the keyboard as if the keyboard was about to disappear, but the keyboard stays up. The same thing happens, when I press the suggestions I get. It seems to only be a problem for account and identity name inout fields for some reason.

https://user-images.githubusercontent.com/76045015/142431527-b71d28ad-cb08-407a-bc38-3f893f330028.MOV

concordium-cl:
Verified in v1.0.26, iOS 14.7.1, iPhone11.

PR #115, opened 2021-11-19
More keyboard fixes

kristiyandobrev:
## Purpose

| iPhone SE  |  iPhone 13 |
|---|---|
|   |  | 

Addresses the last comment in #104 

To begin with, the whole view is very problematic since it has text field, button and a the whole thing is within a scroll view and a keyboard that comes from below. On top of everything we have those huge whitespaces which are not really suitable for smaller screen devices.
Anyway I have applied a fix and removed the scrollview as it was the main culprit of the bug. The scrollview (I believe) was intended to serve smaller screen devices, meaning if the screen is too small then you can scroll the content (not when the keyboard appears). But running it on all different devices proved that the scroll view is totally  unnecessary and therefore I removed it. The result is now “improved” behaviour and the entire content is being pushed up in order to make room for the keyboard. Here from smaller and normal screen devices.

## Changes

* Removed scrollview from `CreateNicknameViewController`
* Using `Combine` instead of `NotificationCenter` 
* Pushing content up when keyboard appears on `CreateNicknameViewController`

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/ConcordiumWallet.xcodeproj/project.pbxproj
+++ b/ConcordiumWallet.xcodeproj/project.pbxproj
@@ -4974,7 +4974,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = "ConcordiumWallet/Resources/Entitlements/Concordium ID.entitlements";
 				CODE_SIGN_STYLE = Automatic;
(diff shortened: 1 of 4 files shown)


td202 (approved):
Looks good.

PR #121, opened 2021-11-24
Prompt user to accept new terms if new terms are available

### LINKED ISSUE #99: Show terms and conditions again after update, if the T&C has changed
**Task description**

At the moment the user is not shown the T&C after an update to the app, that also updated the T&C. 

Each time the T&C has been updated, the user should be shown the new one upon opening the app. However, we might have to let the user make an export, so they don't get locked out if they for some reason don't agree with the new T&C.

ruxandra-constantinescu:
## Purpose

Ask the user to accept new T&C when they are updated
Fixes #99 

## Changes

- saving a hash of the terms and conditions when they are accepted
- compare saved hash of the terms with the hash of the current hash

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


ConcordiumWallet/Settings/AppSettings.swift
```diff
@@ -66,6 +68,62 @@ struct AppSettings {
...
+    + "termsAndConditionsScreen.terms29.title".localized
+    + "termsAndConditionsScreen.terms30.details".localized
+    + "termsAndConditionsScreen.terms31.title".localized
+    + "termsAndConditionsScreen.terms32.details".localized
+    + "termsAndConditionsScreen.terms33.title".localized
+    + "termsAndConditionsScreen.terms34.details".localized
+    + "termsAndConditionsScreen.terms35.title".localized
+    + "termsAndConditionsScreen.terms36.details".localized
+    + "termsAndConditionsScreen.terms37.title".localized
+    + "termsAndConditionsScreen.terms38.details".localized
```
td202, line 116: This seems like it could be annoying to keep updated when the terms change, if sections are added or removed. Would it be possible to combine this somehow with the existing code in `TermsAndConditionsViewController`? Or wouldn't it suffice to have a version number for the terms that is updated with each revision, and just check that?

ConcordiumWallet/Settings/AppSettings.swift
```diff
@@ -66,6 +68,62 @@ struct AppSettings {
...
+    
+    static var termsHash: String? {
```
td202, line 118: Perhaps `acceptedTermsHash` would be more meaningful.

td202 (changes requested):

ConcordiumWallet/Views/AccountsView/AccountsCoordinator.swift
```diff
@@ -14,6 +14,7 @@ protocol AccountsCoordinatorDelegate: AnyObject {
     func createNewIdentity()
     func createNewAccount()
     func noIdentitiesFound()
+    func displayNewTerms()
```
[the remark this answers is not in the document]
ruxandra-constantinescu: Not really :) I made it like this to match the createAccount. But it doesn't need to pass through MenuCoordinator, you are completely right. I just removed that.
I would, however, like to keep both newTermsAvailable and showNewTerms for code clarity

ConcordiumWallet/Settings/AppSettings.swift
ruxandra-constantinescu: The version number would pose the same problem and we agreed with Android that we would do hash, not a version number. I extracted the method that generated the attributed string in a separate file and use that to get the string used for hashing. That should solve any issues regarding sync :)

ConcordiumWallet/Settings/AppSettings.swift
ruxandra-constantinescu: updated :)

td202 (approved):

PR #132, opened 2021-12-03
Stop timer when app resigns active

### LINKED ISSUE #130: App crashes when in background during id issuing that fails
**Bug Description**
After issuing an identity that will fail, I moved away from the app, then returned, then moved away again. App was not closed but in background. When returning for the second time, app crashed.

I also submitted a crash report that might have more info.
 
**Steps to Reproduce**
- issue an identity that will fail
- move app to background
- return to app
- (move through tabs)
- move app to background
- return to app
 
**Expected Result**
App opens and shows failed identity + dialog.

**Actual Result**
App crashed. After restart, app shows failed identity card but no dialog is displayed.

**Versions**
- iOS ccd app v1.0.26
- iPhone11
- iOS 14.7.1

ruxandra-constantinescu:
## Purpose

There seems that there are some issues with Realm objects that are used in timer handlers after app enters background. As per Apple recommendation, timers should be invalidated when app is no longer active ( https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background ). Should fix #130 

## Changes

Added observer for appWillResignActive and appDidBecome active in all three view controllers where timers are used for refresh. The refresh timers for a viewController are stopped when app is deactivated and are scheduled again when app is activated.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/ConcordiumWallet.xcodeproj/project.pbxproj
+++ b/ConcordiumWallet.xcodeproj/project.pbxproj
@@ -4984,7 +4984,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = "ConcordiumWallet/Resources/Entitlements/Concordium ID.entitlements";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 26;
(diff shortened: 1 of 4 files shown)


td202 (approved):
Looks good.

Issue #147, opened 2022-01-14
Backup dlg immediately after upgrading wallet


concordium-cl:
**Description**
After upgrading to a new version, the user should be prompted to backup. Most current users will be in that situation, without necessarily creating an account immediately after upgrading.

Text and buttons:

```
Welcome back!

Did you make a backup of your identities and accounts recently?

If you didn’t, now is a good time to do so! Concordium strongly recommends having a backup of all your identities and accounts. 
```
Left button: Not now
Right button: Make backup

![image](https://user-images.githubusercontent.com/76045015/149350574-51b62754-cdc9-414d-b85d-634db905ddb5.png)

 
**Steps to Reproduce**
- in old version, create id and account
- upgrade to new version

**Expected Result**
There should be an info box asking the user to make a backup, if he didn't do so in the old version, with two buttons "Dismiss" and "Okay" (leading to export). "Dismiss" should lead to the existing warning "are you sure?".

**Actual Result**
Warning dilaogs and "nagging box" first appear after new account creation.

**Versions**
- CCD testnet ?

concordium-cl:
Discovered when testing Android.

kristiyandobrev:
@concordium-cl 

Should that the case also for future updates ?

concordium-cl:
@jens-concordium, what do you think?

jens-concordium:
I guess it should. There could be cases where a user updates from an older version and directly to a later version than this one, in the future. @jensvesti, how is this handled in Android?

concordium-cl:
To be tested

1.0.5 --> newest version: prompt shown
1.1.27 --> newest version: prompt shown

new install of newest version: no prompt

concordium-cl:
1.1.27 --> 1.2.31: prompt shown ✅ 
new install of 1.2.31: no prompt ✅ 
1.0.7 --> 1.2.31: prompt shown ✅ 

Note: Dlg text here is outdated, see https://github.com/Concordium/concordium-reference-wallet-ios/issues/150.

PR #157, opened 2022-01-26
Don't show backup warning on fresh install

kristiyandobrev:
## Purpose

Currently the backup warning alert is shown after a fresh app install, which should not be the case

Relates to #147 

## Changes

* Made `hasRunBefore` a setting in `AppSettings` 
* Checking for `hasRunBefore` before showing the backup warning alert

## Checklist

- [x] The code compiles without warnings.
      hard-to-understand areas.

### DIFF
--- a/ConcordiumWallet/Views/MoreSection/VerifyIdsAndAccounts/SanityChecker.swift
+++ b/ConcordiumWallet/Views/MoreSection/VerifyIdsAndAccounts/SanityChecker.swift
@@ -65,7 +65,7 @@ class SanityChecker {
         switch mode {
         case .automatic:
             if report.count == 0 || AppSettings.ignoreMissingKeysForIdsOrAccountsAtLogin == true {
-                if AppSettings.lastKnownAppVersionSinceBackupWarning == nil {
+                if AppSettings.lastKnownAppVersionSinceBackupWarning == nil, AppSettings.hasRunBefore {
                     showBackupWarningAfterUpdate()
                     AppSettings.lastKnownAppVersionSinceBackupWarning = AppSettings.appVersion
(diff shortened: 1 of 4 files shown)


ConcordiumWallet/Views/MoreSection/VerifyIdsAndAccounts/SanityChecker.swift
```diff
@@ -65,7 +65,7 @@ class SanityChecker {
         switch mode {
         case .automatic:
             if report.count == 0 || AppSettings.ignoreMissingKeysForIdsOrAccountsAtLogin == true {
-                if AppSettings.lastKnownAppVersionSinceBackupWarning == nil {
+                if AppSettings.lastKnownAppVersionSinceBackupWarning == nil, AppSettings.hasRunBefore {
```
ruxandra-constantinescu, line 68: hasRunBefore does not give any info about whether it comes from the login screen or from creating an account / import. Has run before is used only to know whether the app has ever been launched by a user and in case it hasn't clear all data that might be left from a previous install (keychain)
You can do it either by sending a param to the AppCoordinator.showMainTabbar depending on where it is being called from, or add a new app setting which you can set in LoginCoordinator.start based on the value of passwordCreated

ruxandra-constantinescu (changes requested):
It needs to be updated so that the Welcome back si not shown after creating accounts on a fresh install

ConcordiumWallet/Views/MoreSection/VerifyIdsAndAccounts/SanityChecker.swift
kristiyandobrev: Thanks for pointing it. It makes sense. Although I have now fixed it a bit differently.

ruxandra-constantinescu (approved):

PR #269, opened 2022-05-16
Force update functionality

XOOPsoft:
## Purpose

Enable wallet-proxy to inform about or force users to update the app from App Store.

## Changes

Fetch appSettings from wallet-proxy.
Parse data and show alert if needed.

## Checklist

- [X] I have documented my code, in particular the intent of the
      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

- [X] I accept the above linked CLA.

### DIFF
--- a/ConcordiumWallet/Model/AppSettingsResponse.swift
+++ b/ConcordiumWallet/Model/AppSettingsResponse.swift
@@ -1,0 +1,20 @@
+//
+//  AppSettingsResponse.swift
+//  ConcordiumWallet
+//
+//  Created by Lars Christensen on 16/05/2022.
+//  Copyright © 2022 concordium. All rights reserved.
+//
+
+import Foundation
+
+struct AppSettingsResponse: Codable {
+    let status: String
+    let url: String?
+}
+
+enum AppStatusType: String, Codable {
+    case ok
+    case warning
+    case needsUpdate
+}
--- a/ConcordiumWallet/Model/AppSettingsResponse.swift
+++ b/ConcordiumWallet/Model/AppSettingsResponse.swift
@@ -9,11 +9,11 @@
 import Foundation
 
 struct AppSettingsResponse: Codable {
-    let status: String
-    let url: String?
+    let status: AppSettingsStatus
+    let url: URL?
 }
 
-enum AppStatusType: String, Codable {
+enum AppSettingsStatus: String, Codable {
     case ok
     case warning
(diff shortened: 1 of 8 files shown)


ConcordiumWallet/Model/AppSettingsResponse.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Created by Lars Christensen on 16/05/2022.
+//  Copyright © 2022 concordium. All rights reserved.
+//
+
+import Foundation
+
+struct AppSettingsResponse: Codable {
+    let status: String
+    let url: String?
+}
```
Niljak, line 14:
```suggestion
enum AppSettingsStatus: String, Codable {
    case ok
    case warning
    case needsUpdate
}

struct AppSettingsResponse: Codable {
    let status: AppSettingsStatus
    let url: URL?
}
```

I would probably create an enum for the status property and have the url property use the Url type. That way it is automatically handled by codable and we can avoid a nasty unwrapping later in the code.

Niljak (changes requested):

ConcordiumWallet/Views/AccountsView/AppSettingsService.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    let networkManager: NetworkManagerProtocol
+    
+    init(networkManager: NetworkManagerProtocol) {
+        self.networkManager = networkManager
+    }
+    
+    func getAppSettigns(platform: String, version: Int) -> AnyPublisher<AppSettingsResponse, Error> {
+        let url = ApiConstants.appSettings.appending("platform", value: platform).appending("version", value: "\(version)")
+        let request = ResourceRequest(url: url)
```
Niljak, line 26: Resource request handles query parameters for us. There is an optional parameter called "parameters" that is then used as query parameters for the request.

ConcordiumWallet/Model/AppSettingsResponse.swift
XOOPsoft: Thanks, I have changed it.

ConcordiumWallet/Views/AccountsView/AppSettingsService.swift
XOOPsoft: I have changed it to use the parameters and removed the extension that was added.

Niljak (approved):

PR #304, opened 2023-03-03
CI: Add initial verification workflow

bisgardo:
The workflow runs on all commits in a pull request targeting `main` as well as commits to `main` itself.

It builds all of the schemes `StagingNet`, `ProdTestNet`, and `ProdMainNet` and runs tests for scheme `Mock`, all in parallel.

The changes to `ConcordiumWallet.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` were necessary to avoid an error related to schema version 2 being unsupported. It's created by deleting the file and let XCode re-generate it.

The changes to `ConcordiumWallet/Views/AccountsView/AccountDetails/AccountDetailsPresenter.swift` were necessary to avoid an error caused by the typechecker taking too long.

Resolves CBW-786.


alexandardimov (approved):
Looks good.

PR #318, opened 2023-05-23
[CBW-1031] fix app flow when new t&c available password is created

milsawicki:
## Purpose
After upgrading an existing wallet though ​[CBW-910](unreleased) and accepting terms and condition, the user is prompted to create a new password. This would allow an attacker to reset the password on a stolen phone, and is undesired from perspective of user experience.

## Changes
I reasured that after current T&C version response is fetched, we do check if user has a password already set. 
I covered the missing scenario with adequate tests.

[CBW-910]: https://concordium.atlassian.net/browse/CBW-910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ConcordiumWallet/AppCoordinator.swift
+++ b/ConcordiumWallet/AppCoordinator.swift
@@ -15,7 +15,7 @@ class AppCoordinator: NSObject, Coordinator, ShowAlert, RequestPasswordDelegate 
 
     var navigationController: UINavigationController
     let defaultProvider = ServicesProvider.defaultProvider()
-    
+
     private var accountsCoordinator: AccountsCoordinator?
     private var needsAppCheck = true
     private var cancellables: [AnyCancellable] = []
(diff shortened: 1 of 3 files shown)


ConcordiumWallet/AppCoordinator.swift
```diff
@@ -49,25 +49,32 @@ class AppCoordinator: NSObject, Coordinator, ShowAlert, RequestPasswordDelegate
...
-                                                parentCoordinator: self,
-                                                dependencyProvider: defaultProvider)
+        let loginCoordinator = LoginCoordinator(
+            navigationController: navigationController,
+            parentCoordinator: self,
+            dependencyProvider: defaultProvider,
+            termsAndCondtionsFactory: { [weak self] termsAndConditions in
+                guard let self = self else { return nil }
+                let viewModel = TermsAndConditionsViewModel(storageManager: self.defaultProvider.storageManager(), termsAndConditions: termsAndConditions)
+                return viewModel
```
milsawicki, line 59: remove the local var

milsawicki, line 59:
```suggestion
                return TermsAndConditionsViewModel(storageManager: self.defaultProvider.storageManager(), termsAndConditions: termsAndConditions)
```

bisgardo (approved):

ConcordiumWalletTests/Coordinators/LoginCoordinatorTests.swift
```diff
@@ -18,7 +18,7 @@ class LoginCoordinatorTests: XCTestCase {
     var keychainMock: InMemoryKeychain!
     var dependencyProvider: LoginDependencyProviderMock!
     private var cancellables: Set<AnyCancellable>!
-
+    private var tacFactory: TermsAndConditionsViewFactory!
```
bisgardo, line 21: @milsawicki what is the meaning of `!` here?

PR #320, opened 2023-05-25
Improve consistency of testcase naming

bisgardo:


milsawicki (approved):
makes sense - word view is used to express a part of screen 👍🏻

PR #329, opened 2023-06-22
WalletConnect: Implement request handling

milsawicki:
- Extract WalletConnect listeners from `AccountCoordinator` to new `WalletConnectCoordinator` and ensure that it's initialized only once. This ensures that events are handled only once and thus that views are pushed only ones.

- Add decoding for supported `WalletConnect` payloads, bridge code for invoking `parameter_to_json_input` in crypto lib, and support for contract update in transaction builder and storage object.

- Implement request approval/reject logic and add corresponding views.

- Extract account from session data - this isn't reliable enough as it enables use of unexpected account from old session! This includes checks that the only the "ccd" namespace with only a single account (which exists) is provided/defined. We should also check that it has the expected methods and events. If the checks fail we report an error in the app but doesn't actually reject the request. This will be added afterwards.

- Improve debug view from QR scanner.

- CI: Use fixed Xcode version

- `showMainTabbar`: Only reinitialise `AccountsCoordinator` if one isn't already running - is that still relevant given that we explicitly nil it?


ConcordiumWallet/AppCoordinator.swift
```diff
@@ -65,13 +65,15 @@ class AppCoordinator: NSObject, Coordinator, ShowAlert, RequestPasswordDelegate
...
         navigationController.setupBaseNavigationControllerStyle()
 
-        accountsCoordinator = AccountsCoordinator(
-            navigationController: navigationController,
-            dependencyProvider: defaultProvider,
-            appSettingsDelegate: self,
-            accountsPresenterDelegate: self
-        )
-        accountsCoordinator?.start()
+        if accountsCoordinator == nil {
```
bisgardo, line 68: Is that still relevant given that we explicitly set `accountsCoordinator` to `nil` in `logout`?

bisgardo: considering for next PR

bisgardo:
Force merged because the workflow seems indefinitely queued. Tests pass locally.

PR #331, opened 2023-07-03
Improve error handling

bisgardo:
- Extract helper functions for responding with or without errors.
- Reject proposals with unexpected "required" namespaces.
- Define errors as big nested enum.

### DIFF
--- a/ConcordiumWallet/Model/WalletConnectError.swift
+++ b/ConcordiumWallet/Model/WalletConnectError.swift
@@ -1,0 +1,81 @@
+import Foundation
+
+enum SchemaError: Error {
+    case invalidBase64(String)
+    case unknownFormat
+    
+    var codeAndMsg: (Int, String) {
+        switch self {
+        case .invalidBase64(_):
+            return (1110, "Provided schema is not valid base64")
+        case .unknownFormat:
+            return (1120, "Unknown schema format")
+        }
+    }
+}
+
+enum SessionError: Error {
+    case sessionNotFound(topic: String)
+    case unexpectedNamespaces(namespaces: [String])
+    case unexpectedAccountCount(addresses: [String])
+    case accountNotFound(address: String)
+    
+    var codeAndMsg: (Int, String) {
+        switch self {
+        case let .sessionNotFound(topic: topic):
+            return (7001, "Session not found for topic \(topic)")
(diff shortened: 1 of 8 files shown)


ConcordiumWallet/Model/WalletConnectError.swift
```diff
@@ -0,0 +1,81 @@
...
+        case let .unsupportedTransactionType(type):
+            return (8100, "Unsupported transaction type '\(type)' (only 'Update' is supported)")
+        case let .invalidPayload(msg):
+            return (8200, "Invalid payload: \(msg)")
+        }
+    }
+}
+
+enum WalletConnectError: Error {
+    case unknownError // TODO: remove
```
milsawicki, line 52: is it worth keeping that error. I see we don't use that error anywhere, do we?

milsawicki (approved):

ConcordiumWallet/Model/WalletConnectError.swift
bisgardo: Good catch. It's not used, removed it.

ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift
```diff
@@ -92,67 +87,85 @@ private extension WalletConnectCoordinator {
...
+                // We only check required namespaces, not the optional ones.
+                guard proposal.requiredNamespaces.count == 1, let ccdNamespace = proposal.requiredNamespaces[expectedNamespaceKey] else {
+                    self.reject(proposal: proposal, reason: .userRejected, msg: "Unexpected namespaces: \(proposal.requiredNamespaces.keys)", shouldPresent: true)
+                    return
+                }
+                if let chains = ccdNamespace.chains, chains != expectedBlockchains {
+                    self.reject(proposal: proposal, reason: .userRejectedChains, msg: "Expected chain \"\(expectedBlockchains)\" bot got \(chains)", shouldPresent: true)
+                    return
+                }
+                if ccdNamespace.events != expectedEvents {
```
bisgardo, line 108: We need to check that `expectedEvents` contains all `ccdNamespace.events`, not that they're the same. Not all dapps request all methods.

We can still allow them all.

Same with methods below.

PR #344, opened 2023-07-11
Remove election difficulty from expected parameters.

abizjak:
## Purpose

Election difficultly is no longer present in P6, which causes failure of parameter parsing for `GET /v0/chainParameters`.

Since this is not used by the wallet anyhow, it is removed from the expected parameters.

## Checklist

      hard-to-understand areas.


orhoj:
@milsawicki Note that we also need to implement the same change for the legacy wallet.

milsawicki (approved):

PR #346, opened 2023-07-12
Fix concordex schema with fixed-length Base64 format for the string

milsawicki:
## Purpose

Fixes: https://concordium.atlassian.net/browse/CBW-1170
There was a problem where Both Dex and wallet showed error “Converting contract update parameters failed”, refer attached screenshot “Dex_token swap error“. 

## Changes
Added a that ensures that a string representing base64-encoded data has a proper format. If there's string not divadable by 4 it adds proper amount of `=` sings in order to coply with base64 format.


ConcordiumWallet/Extensions/String+Base64.swift
```diff
@@ -0,0 +1,19 @@
...
+
+import Foundation
+
+///      Returns the fixed base64 format of the string.
+/// This method is used to ensure that a string representing base64-encoded data has a proper format. In some cases,
+/// when working with certain schemas or data sources, the base64 string returned may not adhere to the standard format.
+/// This method addresses that issue by adding padding characters ('=') to the end of the string if its length is not a multiple of 4, thus making it a valid base64 format.
+
+extension String {
+    var fixedBase64Format: Self {
```
orhoj, line 14: Ideally there would be unit tests for this method that verifies it doesn't alter base64 strings that already have padding, and some to verify that correct padding is added for those that are missing it.

orhoj (approved):

ConcordiumWallet/Extensions/String+Base64.swift
milsawicki: yes, it's perfect case for tests

PR #368, opened 2023-10-06
WalletConenct deeplink and universal link support

prinshamlet:
Scope of the task was to implement support for WalletConnect deeplinking [CBW-1441](https://concordium.atlassian.net/browse/CBW-1441) and universal links ([CBW-1442](https://concordium.atlassian.net/browse/CBW-1442)). 

[CBW-1441]: https://concordium.atlassian.net/browse/CBW-1441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CBW-1442]: https://concordium.atlassian.net/browse/CBW-1442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Universal link flow demo:
Simulator Screen Recording - iPhone 15 Pro - 2023-10-26 at 21 16 24

Deeplink flow demo:


ConcordiumWallet/AppCoordinator.swift
```diff
@@ -61,6 +61,22 @@ class AppCoordinator: NSObject, Coordinator, ShowAlert, RequestPasswordDelegate
...
+            if let uriQueryItem = queryItems.first(where: { $0.name == "uri" }) {
+                if let uriValue = uriQueryItem.value {
+                    let walletConnectCoordinator = WalletConnectCoordinator(
+                        navigationController: navigationController,
+                        dependencyProvider: self.defaultProvider,
+                        parentCoordiantor: self
+                    )
+                    walletConnectCoordinator.start(with: uriValue)
+                    childCoordinators.append(walletConnectCoordinator)
+                }
```
prinshamlet, line 76: consider guarding it

ConcordiumWallet/AppDelegate.swift
```diff
@@ -92,7 +92,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             receivedCreateIdentityCallback(url)
         } else {
```
prinshamlet, line 93: do extra check

ConcordiumWallet/AppCoordinator.swift
```diff
@@ -62,6 +62,27 @@ class AppCoordinator: NSObject, Coordinator, ShowAlert, RequestPasswordDelegate
...
+        if let uriValue = queryItems.first(where: { $0.name == "uri" })?.value {
+            let walletConnectCoordinator = WalletConnectCoordinator(
+                navigationController: navigationController,
+                dependencyProvider: defaultProvider,
+                parentCoordiantor: self
+            )
+            walletConnectCoordinator.start(with: uriValue)
+            childCoordinators.append(walletConnectCoordinator)
+        }
+        else if let requestId = queryItems.first(where: { $0.name == "requestId" })?.value,
```
orhoj, line 80: What is this used for?

orhoj (changes requested):

ConcordiumWallet/AppCoordinator.swift
prinshamlet: I used it to diffrentiate between session proposal and request deep link for sake of debuging. I think  there's no use for these variables now, because the WalletConnectCoordinator listens for requests anyway. We could use it for checking if the correct request is currently displayed, but it's not super easy, so for now we can just give up on this.

orhoj:
To me the changes look good now, but I don't have an option to test it right now. I would like @bisgardo to check it out and try it before we merge it.

ConcordiumWallet/AppDelegate.swift
```diff
@@ -86,13 +86,25 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    }
+
     func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
         Logger.trace("application:openUrl: \(url)")
         if url.absoluteString.starts(with: ApiConstants.notabeneCallback) {
             receivedCreateIdentityCallback(url)
-        } else {
-            // importing file
-            appCoordinator.importWallet(from: url)
+        } else if url.absoluteString.starts(with: ApiConstants.walletConnectURI) {
```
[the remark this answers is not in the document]
bisgardo: Just verified that the scheme will open the app but not walletconnect.

ConcordiumWallet/Views/WalletConnect/AccountSelect/WalletConnectAccountSelectView.swift
```diff
@@ -20,6 +20,19 @@ struct WalletConnectAccountSelectView: View {
...
+                Button {
+                    viewModel.dismissView()
+                } label: {
+                    HStack {
+                        Image(systemName: "chevron.backward")
+                        Text("Cancel")
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
bisgardo: If I do the same on the current testnet build, I get an error "Pairing with topic (...) already exist". Afterwards the same error occurs, so it isn't actually a new bug.

ConcordiumWallet/AppDelegate.swift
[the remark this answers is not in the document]
prinshamlet: I think it should be 
```    static let walletConnectURI = "\(scheme)://wc"```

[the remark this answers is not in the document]
Comment (bisgardo) [reply ConcordiumWallet/AppDelegate.swift:104]: Testnet needs to support two schemes. You can check against both or use a regex.

File: ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift [hunk 6]
```diff
@@ -182,6 +190,7 @@ private extension WalletConnectCoordinator {
     }
 
     private func setupWalletConnectSettleBinding() {
+
```
bisgardo, line 193: ?

ConcordiumWallet/AppCoordinator.swift
```diff
@@ -62,6 +62,23 @@ class AppCoordinator: NSObject, Coordinator, ShowAlert, RequestPasswordDelegate
...
     }
 
+    func openWalletConnectAccountSelection(url: URL) {
+        guard let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: false)?.queryItems else {
+            showErrorAlert(ViewError.simpleError(localizedReason: "WalletConnect URL format incorrect."))
+            return
+        }
+
+        if let uriValue = queryItems.first(where: { $0.name == "uri" })?.value {
+            let walletConnectCoordinator = WalletConnectCoordinator(
```
bisgardo, line 72: Shouldn't we dismiss any existing WC coordinator first?

ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift
```diff
@@ -21,11 +21,11 @@ class WalletConnectCoordinator: Coordinator {
     init(
         navigationController: UINavigationController,
         dependencyProvider: DependencyProvider,
-        parentCoordiantor: WalletConnectCoordiantorDelegate
+        parentCoordinator: WalletConnectCoordiantorDelegate
     ) {
         self.dependencyProvider = dependencyProvider
         self.navigationController = navigationController
-        parentCoordinator = parentCoordiantor
+        self.parentCoordinator = parentCoordinator
```
bisgardo, line 28: :smile:

ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift
```diff
@@ -156,6 +160,10 @@ private extension WalletConnectCoordinator {
                             ),
                             animated: true
                         )
+                    }, dismissView: {
+                        self.navigationController.popViewController(animated: true)
+                        self.parentCoordinator?.dismissWalletConnectCoordinator()
+                        self.nukeWalletConnectSessionsAndPairings()
```
bisgardo, line 166:
```suggestion
                    }, dismissView: { [weak self] in
                        self?.navigationController.popViewController(animated: true)
                        self?.parentCoordinator?.dismissWalletConnectCoordinator()
                        self?.nukeWalletConnectSessionsAndPairings()
```

bisgardo (commented):
Please add PR description.

ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift
```diff
@@ -605,6 +607,21 @@ extension WalletConnectCoordinator: WalletConnectDelegate {
...
+        Task {
+            do {
+                guard let uri = WalletConnectURI(string: uri) else {
+                    self.presentError(with: "errorAlert.title".localized, message: "Unable to initialize WalletConnectURI.")
+                    return
+                }
+                try await Pair.instance.pair(uri: uri)
+            } catch let err {
+                self.navigationController.popViewController(animated: true)
+                self.presentError(with: "errorAlert.title".localized, message: err.localizedDescription)
```
bisgardo, line 620: missing a dismissal?

ConcordiumWallet/AppCoordinator.swift
```diff
@@ -62,6 +62,27 @@ class AppCoordinator: NSObject, Coordinator, ShowAlert, RequestPasswordDelegate
...
+                let walletConnectCoordinator = WalletConnectCoordinator(
+                    navigationController: navigationController,
+                    dependencyProvider: defaultProvider,
+                    parentCoordinator: self
+                )
+                walletConnectCoordinator.start(with: uriValue)
+                childCoordinators.append(walletConnectCoordinator)
+            }
+        }
+    }
```
bisgardo, line 84: Maybe this is a bit more explicit?

```suggestion
    func openWalletConnectAccountSelection(url: URL) {
        guard let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: false)?.queryItems else {
            showErrorAlert(ViewError.simpleError(localizedReason: "WalletConnect URL format incorrect."))
            return
        }

        if let uriValue = queryItems.first(where: { $0.name == "uri" })?.value {
            getOrCreateWalletConnectCoordinator().start(with: uriValue)
        }
    }
    
    func getOrCreateWalletConnectCoordinator() -> WalletConnectCoordinator {
        if let c = childCoordinators.first(where: { $0 is WalletConnectCoordinator }) as? WalletConnectCoordinator {
            return c
        }
        let c = WalletConnectCoordinator(
            navigationController: navigationController,
            dependencyProvider: defaultProvider,
            parentCoordinator: self
        )
        childCoordinators.append(c)
        return c
    }
```

bisgardo: But what if the user is already in the WC flow? Wouldn't it be better to close that flow first then? I.e. using `dismissWalletConnectCoordinator`.

prinshamlet: hmm that would be the best approach correct

bisgardo: In fact, if there is already a WC connection, we still get the duplicate account list. I then got into a state where the disconnect button no longer worked.

I think connecting via deep link while already connected might be enough of an edge case that we shouldn't worry about it, but it's worth checking if the other solution avoids the issue.

ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift
```diff
@@ -605,6 +607,22 @@ extension WalletConnectCoordinator: WalletConnectDelegate {
         navigationController.pushViewController(vc, animated: true)
     }
 
+    private func pairWalletConnect(with uri: String) {
+        Task {
```
bisgardo, line 610: Intentional to not use weak self?

ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift
prinshamlet: added

ConcordiumWallet/Views/WalletConnect/WalletConnectCoordinator.swift
prinshamlet: overlooked

bisgardo (approved):
Seems to work :tada: !

PR #371, opened 2023-11-07
[CBW-1459] Allow choosing commission rates in configure baker flow

prinshamlet:
## Purpose
Currently the commission rate page in the configure baker flows just displays the rates that will be chosen, regardless of whether the allowed ranges are a singleton or not. We need to make it possible to choose the rate.

## Changes
Simulator Screenshot - iPhone 15 Pro - 2023-11-07 at 01 28 03

Comission rates sliders and textfield has been added in order to allow users to define and modify the values from two screen: 
- Register as baker screen
- Update pool settings

Task: [CBW-1459](https://concordium.atlassian.net/browse/CBW-1459)

[CBW-1459]: https://concordium.atlassian.net/browse/CBW-1459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+            viewModel.error = nil
+        }
+    }
+
+    var body: some View {
+        VStack {
+            if
+                let bakingCommissionRange = viewModel.bakingCommissionRange,
+                let transactionCommissionRange = viewModel.transactionCommissionRange,
+                let finalizationCommissionRange = viewModel.finalizationCommissionRange {
```
bisgardo, line 28: Looks like this should be extracted into a helper method/property. Also, is it intended that if just one of these values are `nil` then you cannot set any of them?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+                        .frame(maxWidth: .infinity)
+                        .foregroundColor(Pallette.whiteText)
+                        .background(Pallette.primary)
+                }
+
+                .background(Pallette.primary)
+                .cornerRadius(10)
+            } else {
+                Spacer()
+                LoadingIndicator()
```
bisgardo, line 62: What are we waiting for here? And how to we move on?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+            // Rounding for edge cases of sliders to prevent odd values caused by floating point errors.
+            if $0 + BakerCommissionSettingsView.sliderStep > range.max || $0 - BakerCommissionSettingsView.sliderStep < range.min {
+                commission = round($0 * 100) / 100
+            } else {
+                self.commission = $0
+            }
+        })
+        HStack {
+            VStack {
+                Text("Min:")
```
bisgardo, line 97: Would look nicer without the colon. And maybe smaller font size. But fine if this is the agreed upon design.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+//  Copyright © 2023 concordium. All rights reserved.
+//
+
+import Combine
+import SwiftUI
+
+struct BakerCommissionSettingsView: View {
+    static var sliderStep = 1e-5
+    @StateObject var viewModel: BakerCommissionSettingsViewModel
+    var isShowingError: Binding<Bool> {
```
bisgardo, line 15: Unused - assuming this was replaced by `isNotNil`?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+                Spacer()
+                LoadingIndicator()
+                Spacer()
+            }
+        }
+        .alert(
+            "Error",
+            isPresented: $viewModel.error.isNotNil(),
+            presenting: viewModel.error,
+            actions: { _ in },
```
bisgardo, line 70: Is there nothing to do on error? What happens then?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+                Button(action: viewModel.continueButtonTapped) {
+                    Text("Continue")
+                        .padding()
+                        .frame(maxWidth: .infinity)
+                        .foregroundColor(Pallette.whiteText)
+                        .background(Pallette.primary)
+                }
+
+                .background(Pallette.primary)
+                .cornerRadius(10)
```
bisgardo, line 59: And this should be its own view.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+            VStack {
+                HStack(alignment: .center, spacing: 1) {
+                    Spacer()
+                    TextField("", value: $commission, formatter: NumberFormatter.comissionFormatter)
+                        .disabled(range.min == range.max && allowCustomCommissionRates)
+                        .keyboardType(.decimalPad)
+                        .textFieldStyle(.roundedBorder)
+                        .lineLimit(1)
+                        .multilineTextAlignment(.center)
+
```
bisgardo, line 109:
```suggestion
```

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+        HStack {
+            VStack {
+                Text("Min:")
+                Text("\(formatter.string(from: NSNumber(value: range.min)) ?? " - ")%")
+            }
+            VStack {
+                HStack(alignment: .center, spacing: 1) {
+                    Spacer()
+                    TextField("", value: $commission, formatter: NumberFormatter.comissionFormatter)
+                        .disabled(range.min == range.max && allowCustomCommissionRates)
```
bisgardo, line 104: so if `allowCustomCommissionRates` is false then the field is enabled?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+            VStack {
+                Text("Max:")
+                Text("\(formatter.string(from: NSNumber(value: range.max)) ?? " - ")%")
+            }
+        }
+        .padding(.bottom, 16)
+    }
+}
+
+#if DEBUG
```
bisgardo, line 129: Doesn't look like something that should be kept?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+                        .multilineTextAlignment(.center)
+
+                        .frame(width: 72)
+                    Text(" %")
+                    Spacer()
+                }
+                // `ALLOW_CUSTOM_COMMISSION_RATES` flag can be added in arguments passed on launch in scheme settings.
+                // Used for toggling on/off slider feature.
+                if range.min < range.max && allowCustomCommissionRates {
+                    Slider(value: commissionBinding, in: range.min ... range.max, step: BakerCommissionSettingsView.sliderStep)
```
bisgardo, line 117: Didn't we decide to set step to `1.0`?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+    @Binding var commission: Double
+    let formatter: NumberFormatter = .comissionFormatter
+    private let allowCustomCommissionRates = UserDefaults.standard.bool(forKey: "ALLOW_CUSTOM_COMMISSION_RATES")
+    var body: some View {
+        let commissionBinding = Binding<Double>(get: {
+            self.commission
+        }, set: {
+            // Rounding for edge cases of sliders to prevent odd values caused by floating point errors.
+            if $0 + BakerCommissionSettingsView.sliderStep > range.max || $0 - BakerCommissionSettingsView.sliderStep < range.min {
+                commission = round($0 * 100) / 100
```
bisgardo, line 90: Why not just clamp the value to the range?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,227 @@
...
+        .onAppear { viewModel.fetchData() }
+        .padding()
+    }
+}
+
+struct BakerCommissionSliderView: View {
+    var range: CommissionRange
+    @Binding var commission: Double
+    let formatter: NumberFormatter = .comissionFormatter
+    private let allowCustomCommissionRates = UserDefaults.standard.bool(forKey: "ALLOW_CUSTOM_COMMISSION_RATES")
```
bisgardo, line 83: I don't understand this field `allowCustomCommissionRates`. It only controls whether the input field is disabled (which it is only if this value is *true*) and whether to show the slider.

Also why is it a build parameter? When would you set this to false? The feature is controlled by the chain parameters anyway.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,142 @@
...
+}
+
+class BakerCommissionSettingsViewModel: ObservableObject {
+    enum BakerCommissionSettingError: LocalizedError {
+        case transactionFeeOutOfRange
+        case bakingRewardOutOfRange
+        case finalizationRewardOutOfRange
+        case networkError(Error)
+
+        var errorMessage: String? {
```
bisgardo, line 29:
```suggestion
        var errorMessage: String {
```

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: Acutally I wasn't sure what final conclusion was was since there were no reactions from the team. Now it's the same way as on Android. If you think it can be done 'our' way and won't be pointed out as /bug/ then I am all for it :)

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: yes, sorry for that

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: yes, these values are brought from API and are required. If there won't be at least of them returned it means that's something is wrong with the API. Any of these values missing might destroy/cause errors on further steps like updating or setting up baking transaction.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: We are waiting for results from line 67 ` .onAppear { viewModel.fetchData() }` to fetch commission range. We don't want users to be able to proceed without those, do we?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: we just dismiss the alert view, that's why i left it empty. No coloring/further actions are done. On Android it only shows toast view on the bottom, so I think it corresponds with it well.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: which part exactly you mean? :)

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: better

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: It was used to prevent displaying slider/ textfield editing values on testnet and mainnet as the feature was only working on stagenet. I didn't know when it will make it to testnet at that point.  I added this mechanism to allow merging it before the backend release. However, it was announced today that the release will happen this week, so I am removing this flag now

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: I copy Andorid 🤷🏻‍♂️ anytime I try to improve something it returns as bug 😄

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: explained above.  should be disabled when range is 0 and doesnt allow custom rates (`!allowCustomCommissionRates`). My bad, but luckly the flag is not required now :)

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: this is compiler flag so we dont compile it for relaase builds. It's just my whim 😄 we might aswell remove preview as I used it for building the view, remove the flag or leave it. doesn't make much difference

ConcordiumWallet/Extensions/Binding.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+
+import SwiftUI
+
+extension Binding {
+    func isNotNil<T>() -> Binding<Bool> where Value == T? {
+        .init(get: {
+            wrappedValue != nil
+        }, set: { _ in
+            wrappedValue = nil
```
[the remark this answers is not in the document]
prinshamlet: yes. in theory it should not happen as as you should not be allowed to assign this value. ignoring it would be better 👍🏻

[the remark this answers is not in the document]
prinshamlet: btw. it's a bummer that apple doesn't provide something to make this kind of check

ConcordiumWallet/Extensions/Binding.swift
```diff
@@ -0,0 +1,19 @@
...
+//  ConcordiumWallet
+//
+//  Created by Milan Sawicki on 07/11/2023.
+//  Copyright © 2023 concordium. All rights reserved.
+//
+
+import SwiftUI
+
+extension Binding {
+    func isNotNil<T>() -> Binding<Bool> where Value == T? {
```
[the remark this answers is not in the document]
prinshamlet: done

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: It would be nice with a converter
```
(bakingCommissionRange?, transactionCommissionRange?, finalizationCommissionRange?) -> (bakingCommissionRange, transactionCommissionRange, finalizationCommissionRange)?
```
i.e. collecting the unwrapped values into a single nillable object with non-nil fields. Then the `if` would only test whether that value is nil.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: the part made up by the lines covered by the comment (22 to 52) :)

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: Ok that's fair but if the API responds with something unexpected you'll be stuck with this indicator. So you need to add some error handling. For example the converter I mentioned above could return a `Result?` which is `nil` until the API result is ready and `.success` only if the reply is valid.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: ok

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: Let's not risk that then ;)

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: I know but it's kind of like keeping dead code IMO. Unless you have some specific use of it in mind going forward?

prinshamlet: not really, only if we wanted to modify this. but can be easly recreated

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: changed

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: > Ok that's fair but if the API responds with something unexpected you'll be stuck with this indicator. So you need to add some error handling. For example the converter I mentioned above could return a `Result?` which is `nil` until the API result is ready and `.success` only if the reply is valid.

there's optional error that triggers alert display. (func fetchData in viewModel)

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,116 @@
...
+                )
+                Spacer()
+                Button(action: viewModel.continueButtonTapped) {
+                    Text("Continue")
+                        .padding()
+                        .frame(maxWidth: .infinity)
+                        .foregroundColor(Pallette.whiteText)
+                        .background(Pallette.primary)
+                }
+
```
prinshamlet, line 47: I dont feel like this needs larger extraction to be honest. do you still insist?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,116 @@
...
+                    Text(" %")
+                    Spacer()
+                }
+                if range.min < range.max {
+                    Slider(value: commissionBinding, in: range.min ... range.max, step: BakerCommissionSettingsView.sliderStep)
+                }
+            }
+            VStack {
+                Text("Max:")
+                Text("\(formatter.string(from: NSNumber(value: range.max)) ?? " - ")%")
```
bisgardo, line 111: How could `formatter.string` return `nil`? If never, maybe we should just force unwrap? Displaying ` - %` doesn't seem very informative.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+//
+//  Created by Milan Sawicki on 06/11/2023.
+//  Copyright © 2023 concordium. All rights reserved.
+//
+
+import Combine
+import Foundation
+
+extension NumberFormatter {
+    static var comissionFormatter: NumberFormatter {
```
bisgardo, line 13:
```suggestion
    static var commissionFormatter: NumberFormatter {
```

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+        finalizationCommissionRange: CommissionRange
+    )?
+    @Published var error: BakerCommissionSettingError?
+
+    private var cancellables = Set<AnyCancellable>()
+    private var didTapContinue: () -> Void
+    private var service: StakeServiceProtocol
+    private var handler: StakeDataHandler
+
+    private static let commisionMultiplier = 100
```
bisgardo, line 58:
```suggestion
    private static let commissionMultiplier = 100
```

ConcordiumWallet/Views/Stake/Common/StakeDataHandler.swift
```diff
@@ -356,13 +356,28 @@ struct RestakeBakerData: SimpleFieldValue {
     }
 }
 
-struct BakerComissionData: FieldValue {
+struct BakerComissionData: FieldValue, Equatable {
```
bisgardo, line 359:
```suggestion
struct BakerCommissionData: FieldValue, Equatable {
```

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+        guard ranges.finalizationCommissionRange.min ... ranges.finalizationCommissionRange.max ~= finalizationRewardCommission else {
+            throw BakerCommissionSettingError.finalizationRewardOutOfRange
+        }
+
+        guard ranges.transactionCommissionRange.min ... ranges.transactionCommissionRange.max ~= transactionFeeCommission else {
+            throw BakerCommissionSettingError.transactionFeeOutOfRange
+        }
+    }
+
+    private func updateCommisionValues(baking: Double, transaction: Double, finalization: Double) {
```
bisgardo, line 136:
```suggestion
    private func updateCommissionValues(baking: Double, transaction: Double, finalization: Double) {
```

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+        .store(in: &cancellables)
+    }
+
+    func continueButtonTapped() {
+        do {
+            try validate()
+            handler.add(entry: BakerComissionData(bakingRewardComission: bakingRewardCommission, finalizationRewardComission: finalizationRewardCommission, transactionComission: transactionFeeCommission))
+            didTapContinue()
+        } catch let error {
+            self.error = error as? BakerCommissionSettingError
```
bisgardo, line 114: So if the error isn't a `BakerCommissionSettingError` we fail silently?

ConcordiumWallet/Views/Stake/Common/StakeDataHandler.swift
```diff
@@ -356,13 +356,28 @@ struct RestakeBakerData: SimpleFieldValue {
...
     let field = Field.bakerComission
```
bisgardo, line 360:
```suggestion
    let field = Field.bakerCommission
```

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+                    return
+                }
+                if let data = self.handler.getCurrentEntry(BakerComissionData.self) {
+                    self.updateCommisionValues(
+                        baking: data.bakingRewardComission,
+                        transaction: data.transactionComission,
+                        finalization: data.finalizationRewardComission
+                    )
+                    return
+                }
```
bisgardo, line 89: I don't understand what's happening here. Please add some comments explaining it.

At least: What is new/current entry? Why is this `handler` invoked in response to a successful result but not using the `response` object at all? Why is `self.commissionRanges` not updated in these cases?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Published var transactionFeeCommission: Double = 0
+    @Published var finalizationRewardCommission: Double = 0
+    @Published var bakingRewardCommission: Double = 0
+    @Published var commissionRanges: (
+        bakingCommissionRange: CommissionRange,
+        transactionCommissionRange: CommissionRange,
+        finalizationCommissionRange: CommissionRange
+    )?
```
bisgardo, line 50: Why not a struct?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+        do {
+            try validate()
+            handler.add(entry: BakerComissionData(bakingRewardComission: bakingRewardCommission, finalizationRewardComission: finalizationRewardCommission, transactionComission: transactionFeeCommission))
+            didTapContinue()
+        } catch let error {
+            self.error = error as? BakerCommissionSettingError
+        }
+    }
+
+    func validate() throws {
```
bisgardo, line 118: Maybe returning error is a better fit than throwing here? Better type safety.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
bisgardo: Would not be necessary with the suggestion below.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: It looks to me like any kind of error (including parsing error) will be classified as a "network" error?

And that if that happens, you'd see the spinner and alert box. When you dismiss that alert you still see the spinner? What then?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
bisgardo: I see now that with the tuple fields being named this doesn't really matter.

prinshamlet: yep, it's pretty much the same

prinshamlet: can I reslove?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: it might if nsnumber is not a number. for instance when range.max value is not a number as string

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,116 @@
...
+            self.commission
+        }, set: {
+            // Rounding for edge cases of sliders to prevent odd values caused by floating point errors.
+            if $0 + BakerCommissionSettingsView.sliderStep >= range.max {
+                commission = range.max
+            } else if $0 - BakerCommissionSettingsView.sliderStep <= range.min {
+                commission = range.min
+            } else {
+                self.commission = $0
+            }
```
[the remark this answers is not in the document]
prinshamlet: won't argue

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,116 @@
...
+import SwiftUI
+
+struct BakerCommissionSettingsView: View {
+    static var sliderStep = 1e-3
+    @StateObject var viewModel: BakerCommissionSettingsViewModel
+
+    var body: some View {
+        VStack {
+            if let ranges = viewModel.commissionRanges {
+                Text("When you open your baker as a pool, you earn\ncommissions of stake delegated to your pool from another accounts")
```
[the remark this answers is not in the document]
prinshamlet: doesnt make much difference here and I missed that

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
prinshamlet: I liked the syntax of try catch more but fine, no problem

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
prinshamlet: yep, that would be the case. I remade it so if we get network error we pop view on ok button.
`            actions: { error in
                Button {
                    switch error {
                    case .networkError:
                        viewModel.dismissView()
                    default: break
                    }
                    viewModel.error = nil

                } label: {
                    Text("OK")
                }
            },`

BakerCommissionSettingsView from line line 60

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
bisgardo: If you mean like `range.max = NaN` then I think force unwrap is in order as nothing is going to work in that case. Failing silently is the worst we can do IMHO.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    func fetchData() {
+        service.getChainParameters().asResult().sink { result in
+            switch result {
+            case let .success(response):
+
+                self.commissionRanges = (bakingCommissionRange: response.bakingCommissionRange,
+                                         transactionCommissionRange: response.transactionCommissionRange,
+                                         finalizationCommissionRange: response.finalizationCommissionRange
+                )
```
bisgardo, line 78:
```suggestion
                self.commissionRanges = (
                    bakingCommissionRange: response.bakingCommissionRange,
                    transactionCommissionRange: response.transactionCommissionRange,
                    finalizationCommissionRange: response.finalizationCommissionRange
                )
```
Isn't that the style we usually prefer?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
bisgardo: Still missing this.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsView.swift
```diff
@@ -0,0 +1,130 @@
...
+        }, set: {
+            var commission = $0
+            // Snap to the range limits to prevent odd floating point roundoff.
+            if commission > range.max - BakerCommissionSettingsView.sliderStep {
+                commission = range.max
+            }
+            if commission < range.min + BakerCommissionSettingsView.sliderStep {
+                commission = range.min
+            }
+            self.commission = commission.rounded(.toNearestOrAwayFromZero, decimals: 5)
```
bisgardo, line 99:
```suggestion
            self.commission = commission.rounded(.toNearestOrAwayFromZero, decimals: 3)
```

As `commissionFormatter` only shows 3 digits, any digits following that will be just noise?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,155 @@
...
+
+import Combine
+import Foundation
+
+extension NumberFormatter {
+    static var commissionFormatter: NumberFormatter {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .decimal
+        formatter.multiplier = 100
+        formatter.maximumFractionDigits = 3
```
bisgardo, line 17:
```suggestion
        formatter.minimumFractionDigits = 3
        formatter.maximumFractionDigits = 3
```

Wouldn't that give a more stable appearance?

bisgardo (approved):

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
bisgardo: Actually the two last digits is always 0. So I think we should show just one digit?

PR #374, opened 2023-11-09
[CBW-1471] Display commission rates in configure baker flow in legacy Wallet

prinshamlet:
## Purpose
Screen displaying commission rates, between pool status and metadata url in the
- Register as baker
- Update pool settings

Task: https://concordium.atlassian.net/browse/CBW-1471
## Changes
xDˣᴰ


bisgardo:
Just tried with stagenet app and got error
image

Reward commission: 66.6
Finalization reward commission: 100
Transaction commission: 58.2

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,156 @@
...
+                // In that case there are no current values, all data is considered 'new'.
+                // That field is first set in `BakerAmountInputPresenter.loadPoolParameters()`.
+                if let data = self.handler.getNewEntry(BakerCommissionData.self) {
+                    self.updateCommissionValues(
+                        baking: data.bakingRewardComission,
+                        transaction: data.transactionComission,
+                        finalization: data.finalizationRewardComission
+                    )
+                    return
+                }
```
bisgardo, line 92: Wasn't it the intent that this should be "fixed"? It seems like this sets them to "max" (as these are the values set in `loadPoolParameters`)?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,156 @@
...
+                // In order to update UI we set slider values to current values.
+                // When slider is moved, the updated is returned by `handler.getNewEntry`.
+                if let data = self.handler.getCurrentEntry(BakerCommissionData.self) {
+                    self.updateCommissionValues(
+                        baking: data.bakingRewardComission,
+                        transaction: data.transactionComission,
+                        finalization: data.finalizationRewardComission
+                    )
+                    return
+                }
```
bisgardo, line 103: There's no slider, so how could this branch be entered?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,156 @@
...
+            return BakerCommissionSettingError.bakingRewardOutOfRange
+        }
+
+        guard ranges.finalizationCommissionRange.min ... ranges.finalizationCommissionRange.max ~= finalizationRewardCommission else {
+            return BakerCommissionSettingError.finalizationRewardOutOfRange
+        }
+
+        guard ranges.transactionCommissionRange.min ... ranges.transactionCommissionRange.max ~= transactionFeeCommission else {
+            return BakerCommissionSettingError.transactionFeeOutOfRange
+        }
```
bisgardo, line 147: Looks redundant as we're constructing the values to be within range?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,156 @@
...
+        service.getChainParameters()
+            .asResult().sink { result in
+            switch result {
+            case let .success(response):
+
+                self.commissionRanges = (
+                    bakingCommissionRange: response.bakingCommissionRange,
+                    transactionCommissionRange: response.transactionCommissionRange,
+                    finalizationCommissionRange: response.finalizationCommissionRange
+                )
```
bisgardo, line 80: What are the ranges used for when there are no sliders?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,156 @@
...
+                        transaction: data.transactionComission,
+                        finalization: data.finalizationRewardComission
+                    )
+                    return
+                }
+                self.updateCommissionValues(
+                    baking: response.bakingCommissionRange.max,
+                    transaction: response.transactionCommissionRange.max,
+                    finalization: response.finalizationCommissionRange.max
+                )
```
bisgardo, line 108: According to the task, the values should be "min + (max-min) * 0.1".

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,156 @@
...
+        didTapContinue: @escaping (() -> Void),
+        dismissView: @escaping (() -> Void)
+    ) {
+        self.service = service
+        self.didTapContinue = didTapContinue
+        self.handler = handler
+        self.dismissView = dismissView
+    }
+
+    func fetchData() {
```
bisgardo, line 70: It's unclear to me why this is needed. We already fetched the chain params in `loadPoolParameters`. Why not use that result here?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
prinshamlet: when you initialize validator for the first time, then it's considered "new".

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
prinshamlet: yes, left it as a safe guard in case anything unexepcted happened. do you think it's too much?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
prinshamlet: yes, but min/max values are not passed further, so I re-fetch them

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
prinshamlet: I thought these values are fetched from backend and set from backend so no calculations are needed

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
prinshamlet: i think we can actually remove this.

ConcordiumWallet/Views/Stake/Baking/BakingCoordinator.swift
```diff
@@ -110,7 +111,21 @@ class BakingCoordinator: Coordinator {
...
         navigationController.pushViewController(vc, animated: true)
     }
-    
+
+    func showComissionSettings(dataHandler: StakeDataHandler) {
+        let viewModel = BakerCommissionSettingsViewModel(
+            service: dependencyProvider.stakeService(),
+            handler: dataHandler,
+            didTapContinue: { [weak self] in
+            self?.showMetadataUrl(dataHandler: dataHandler)
```
bisgardo, line 120: indentation

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
bisgardo: Why not compute the correct value in `loadPoolParameters` then?

ConcordiumWallet/Views/Stake/Baking/BakingCoordinator.swift
```diff
@@ -110,7 +111,21 @@ class BakingCoordinator: Coordinator {
...
+            }) { [weak self] in
+                self?.navigationController.popViewController(animated: true)
+            }
```
bisgardo, line 123: I prefer providing the closure as a normal named argument as it isn't clear from the context what function is it. Unless this is a standard pattern that we always use (i.e. that the trailing closure of view models is "dismiss"), then it's ok.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+class BakerCommissionSettingsViewModel: ObservableObject {
+    @Published var transactionFeeCommission: Double = 0
+    @Published var finalizationRewardCommission: Double = 0
+    @Published var bakingRewardCommission: Double = 0
+
+    @Published var error: BakerCommissionSettingError?
+    var dismissView: () -> Void
```
bisgardo, line 48: Is this ever called?

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+            )
+        )
+        didTapContinue()
+    }
+
+    private func updateCommissionValues(baking: Double, transaction: Double, finalization: Double) {
+        transactionFeeCommission = transaction
+        bakingRewardCommission = baking
+        finalizationRewardCommission = finalization
+    }
```
bisgardo, line 99: This is only called from one place. Consider inlining.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+        descriptionTextView.isScrollEnabled = false
+    }
+
+    private func setupBindings() {
+        viewModel.$transactionFeeCommission
+            .compactMap { [weak self] in
+                guard let self = self else { return nil }
+                return "\(self.formatter.string(from: NSNumber(value: $0))!)%"
+            }
+            .assign(to: \.text, on: transactionFeeAmountLabel)
```
bisgardo, line 85: I hope the fact that this is set asynchronously doesn't cause any problems with the `transactionFeeAmountLabel` being defined as forced non-nil. 🤔

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+        formatter.multiplier = 100
+        formatter.maximumFractionDigits = 3
+        return formatter
+    }
+}
+
+enum BakerCommissionSettingError: LocalizedError {
+    case transactionFeeOutOfRange
+    case bakingRewardOutOfRange
+    case finalizationRewardOutOfRange
```
bisgardo, line 25:
```suggestion
```

bisgardo (approved):
Looks reasonable! Though I haven't been able to test it.

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewController.swift
prinshamlet: setup bindings is called in viewDidLoad method, so it means all views are already initialized. If the view was unable to initialize at that phase we would've gotten a crash already. so, yeah it's ok

prinshamlet: as Apple claims

>  Called after the view has been loaded. For view controllers created in code, this is after -loadView. For view controllers unarchived from a nib, this is after the view is set.

bisgardo: My concern is whether the publisher assigns the value in a different thread, meaning that it can be assigned arbitrarily long after `viewDidLoad` has completed. I might be wrong about that...

Is your claim is that it's fine that
```
@IBOutlet var transactionFeeAmountLabel: UILabel!
```
is uninitialized because it would have failed before `viewDidLoad` even ran otherwise?

prinshamlet: yep. This can happen if, for some reason, the outlet is not connected properly in the interface builder, but it would then crash before viewDidLoad was ever called. 
for safety we can just do something like this
`   if let label = transactionFeeAmountLabel {
            viewModel.$transactionFeeCommission
                .compactMap { [weak self] in
                    guard let self = self else { return nil }
                    return "\(self.formatter.string(from: NSNumber(value: $0))!)%"
                }
                .assign(to: \.text, on: label)
                .store(in: &cancellables)
        }
        if let label = bakingRewardFeeLabel {
            viewModel.$bakingRewardCommission
                .compactMap { [weak self] in
                    guard let self = self else { return nil }
                    return "\(self.formatter.string(from: NSNumber(value: $0))!)%"
                }
                .assign(to: \.text, on: label)
                .store(in: &cancellables)
        }`

ConcordiumWallet/Views/Stake/Baking/BakingCoordinator.swift
prinshamlet: I think it's actually right. autoformatter also doesn't complain

ConcordiumWallet/Views/Stake/Baking/BakingCoordinator.swift
prinshamlet: i get you. i was alwayes corrected for that but other colleges but I don't mind explicit arugment neither 😄

ConcordiumWallet/Views/Stake/Baking/ComissionSettings/BakerCommissionSettingsViewModel.swift
prinshamlet: that wouldn't hurt actually

PR #388, opened 2024-02-07
release 1.5.0

ramakser:
# Release 1.5.0 relates fixes

## Fixes:
CBW-1640 Shield CCDs with the new account

CBW - 1638 Add proper CIS-2 token image preview: add thumbnail and display images due to [docs](https://proposals.concordium.software/CIS/cis-2.html#token-metadata-json)

CBW-1478 Fix CIS-2 token local db persistency 

CBW-1626  Fix restaking logic, in case when user want update options without changing stake amount

CBW-1643 Fix contract update tx cost refresh logic

### DIFF
--- a/ConcordiumWallet/Model/CIS2/CIS2TokenMetadataDetails.swift
+++ b/ConcordiumWallet/Model/CIS2/CIS2TokenMetadataDetails.swift
@@ -9,6 +9,7 @@ struct CIS2TokenMetadataDetails: Codable, Hashable {
     let decimals: Int?
     let description: String
     let thumbnail: ImageData?
+    let display: ImageData?
     let unique: Bool?
     
     struct ImageData: Codable, Hashable, Equatable {
@@ -26,6 +27,11 @@ struct CIS2TokenMetadataDetails: Codable, Hashable {
         } else {
             self.thumbnail = nil
         }
+        if let urlString = entity.display, let url = URL(string: urlString) {
+            self.display = .init(url: url)
+        } else {
+            self.display = nil
+        }
     }
 }
 
@@ -36,6 +42,7 @@ final class CIS2TokenMetadataDetailsEntity: Object {
     @Persisted var decimals: Int? = nil
     @Persisted var metadataDescription: String = ""
     @Persisted var thumbnail: String? = nil
+    @Persisted var display: String? = nil
     @Persisted var unique: Bool? = nil
     convenience init(
         with metadata: CIS2TokenMetadataDetails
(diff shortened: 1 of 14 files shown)


Radiokot (approved):

ConcordiumWallet/Views/Stake/Baking/AmountInput/BakerAmountInputPresenter.swift
```diff
@@ -149,7 +149,13 @@ class BakerAmountInputPresenter: StakeAmountInputPresenterProtocol {
...
                 return .failure(StakeError.internalError)
             }
-            
+        
+            if let stakedAmount = account.baker?.stakedAmount  {
+                if amount == GTU(intValue: stakedAmount) {
+                    return .success(amount)
+                }
+            }
+
```
bisgardo, line 158:
```suggestion
            if let baker = account.baker, baker.stakedAmount == amount.intValue {
                return .success(amount)
            }
```
is slightly cleaner IMHO.

ConcordiumWallet/Model/CIS2/CIS2TokenMetadataDetails.swift
```diff
@@ -26,6 +27,11 @@ struct CIS2TokenMetadataDetails: Codable, Hashable {
         } else {
             self.thumbnail = nil
         }
+        if let urlString = entity.display, let url = URL(string: urlString) {
+            self.display = .init(url: url)
+        } else {
+            self.display = nil
```
bisgardo, line 33: Seems redundant to assign `nil` to something that's already `nil`?

ConcordiumWallet/Views/Stake/Baking/AmountInput/BakerAmountInputPresenter.swift
bisgardo: Also consider adding a comment explaining why short circuit is done.

bisgardo (approved):

bisgardo:
I recommend doing one fix per PR (unless they're closely related). And naming the PR by the actual changes, not the version to be released. The description should also explain the changes (high level), not just reference the tasks (that requires you to do extra work when auditing, but most importantly, Jira is not public so outsiders have no way to follow). Btw. some of the task links in the current description reference Jira tasks, others Git commits.

Doing that is a bit more work, but will give you better reviews, better ability to bisect bugs later on, and better ability to understand why changes were done the way they were in the past.

ramakser:
> I recommend doing one fix per PR (unless they're closely related). And naming the PR by the actual changes, not the version to be released. The description should also explain the changes (high level), not just reference the tasks (that requires you to do extra work when auditing, but most importantly, Jira is not public so outsiders have no way to follow). Btw. some of the task links in the current description reference Jira tasks, others Git commits.
> 
> Doing that is a bit more work, but will give you better reviews, better ability to bisect bugs later on, and better ability to understand why changes were done the way they were in the past.

100% true.  most fixes was done in send funds flow, so I have combine them in one pr.  and then I have merged them with cis2 fixes, so I have no other choice rather than create abstract release brach 😔

next pr's will be as you write before for sure 🥹

PR #389, opened 2024-02-28
Make more lenient CIS-2 Lookup

ramakser:
## Purpose
- More lenient CIS-2 Lookup. 
- Fix CIS-2 token preview in case when thumbnail data is missed
- Fix token fetch logic for the smart contracts with large amount of tokens. Handle broken token metadata. 

## Changes

- Add paginated call for 'CIS2TokenSelectView'  to show list of the tokens. 
- Add new API's v1 for `v1/CIS2TokenMetadata` and `v1/CIS2TokenBalance`
- Fixed corrupted or missed token metadata handle logic.  Filter out broken tokens from the list
- Fixed thumbnails for CIS-2 in case when thumbnail is no available in metadata  - use display image instead. [Docs](https://proposals.concordium.software/CIS/cis-2.html#token-metadata-json)

### DIFF
--- a/ConcordiumWallet/Views/AccountsView/AccountDetails/TokenSelectionView/CIS2TokenSelectViewModel.swift
+++ b/ConcordiumWallet/Views/AccountsView/AccountDetails/TokenSelectionView/CIS2TokenSelectViewModel.swift
@@ -72,8 +72,6 @@ final class CIS2TokenSelectViewModel: ObservableObject {
                 }
                 
                 await MainActor.run {
-                    isLoading = false
-                    currentPage += 1
                     hasMore = !metadata.metadata.isEmpty
                     
                     if currentPage == 1 {
@@ -81,6 +79,8 @@ final class CIS2TokenSelectViewModel: ObservableObject {
                     } else {
                         tokens += representables
                     }
+                    currentPage += 1
+                    isLoading = false
                 }
             } catch {
                 await MainActor.run {
--- a/ConcordiumWallet/Views/AccountsView/AccountDetails/TokenLookup/CIS2TokensCoordinator.swift
+++ b/ConcordiumWallet/Views/AccountsView/AccountDetails/TokenLookup/CIS2TokensCoordinator.swift
@@ -23,10 +23,12 @@ class CIS2TokensCoordinator: Coordinator, AccountAddressQRCoordinatorDelegate {
     }
 
     func start() {
-        var view = TokenLookupView(service: dependencyProvider.cis2Service(), account: account)
-        view.displayContractTokens = { [weak self] tokens, contractIndex in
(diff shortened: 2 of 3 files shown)


ConcordiumWallet/Views/AccountsView/AccountDetails/TokenLookup/CIS2Service.swift
```diff
@@ -100,17 +100,11 @@ class CIS2Service: CIS2ServiceProtocol {
                 url: ApiConstants.cis2Tokens
                     .appendingPathComponent(contractIndex)
                     .appendingPathComponent(contractSubindex),
-                parameters: ["limit" : "1000"]
+                parameters: ["limit" : "20"]
```
ramakser, line 103: we probably should add here a paginated call, due to current API v1 returns max 20 items in response

Radiokot (approved):

ConcordiumWallet/Views/AccountsView/AccountDetails/TokenSelectionView/CIS2TokenSelectViewModel.swift
```diff
@@ -0,0 +1,127 @@
...
+                    if currentPage == 1 {
+                        tokens = representables
+                    } else {
+                        tokens += representables
+                    }
+                }
+            } catch {
+                await MainActor.run {
+                    isLoading = false
+                }
```
bisgardo, line 88: Consider doing this in a `defer` instead.

ConcordiumWallet/Views/AccountsView/AccountDetails/TokenSelectionView/CIS2TokenSelectViewModel.swift
```diff
@@ -0,0 +1,127 @@
...
+            }
+            
+            for try await data in group {
+                allData.append(data)
+            }
+        }
+        
+        return allData.compactMap { (metadataItem, tokenMetadataDetails) -> (CIS2TokensMetadataItem, CIS2TokenMetadataDetails)? in
+            guard let tokenMetadataDetails = tokenMetadataDetails else { return nil }
```
bisgardo, line 123:
```suggestion
            guard let tokenMetadataDetails else { return nil }
```

ramakser: I need this unwrapped value ` -> (CIS2TokensMetadataItem, CIS2TokenMetadataDetails)`. basically here we filter out  all tokens with empty or broken metadata, and return pair for rest

bisgardo: The suggested change is just syntactic sugar, the semantics is the same.

PR #391, opened 2024-03-11
Contract update failed insufficient energy fix

ramakser:
## Purpose

Estimation transaction fee is not always the same for the same transaction.
Fix CIS2 token look list image

## Changes

Fix transaction fee update logic while user input amount of token to send. 
Replaced SDWebImage with Async image due to hight memory consumption during list tokens scroll

### DIFF
--- a/ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
+++ b/ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
@@ -336,6 +336,10 @@ class SendFundPresenter: SendFundPresenterProtocol {
             .store(in: &cancellables)
         
         view.amountTextPublisher
+            .map { [weak self] input in
+                self?.isUpdatingTransferCost = true
+                return input
+            }
             .debounce(for: 0.7, scheduler: DispatchQueue.main)
             .sink{ [weak self] _ in
                 self?.updateTransferCostEstimate()
@@ -556,8 +560,6 @@ class SendFundPresenter: SendFundPresenterProtocol {
     }
 
     private func updateTransferCostEstimate() {
-        isUpdatingTransferCost = true
-        
         dependencyProvider
             .transactionsService()
             .getTransferCost(transferType: transferType.actualType.toWalletProxyTransferType(), costParameters: buildTransferCostParameter())
--- a/ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
+++ b/ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
@@ -557,6 +557,8 @@ class SendFundPresenter: SendFundPresenterProtocol {
                 self?.energy = value.energy
                 let feeMessage = "sendFund.feeMessage".localized + cost.displayValue()
                 self?.viewModel.feeMessage = feeMessage
+                
+                debugPrint("Update Transfer Cost Estimates -- \(value)")
             })
             .store(in: &cancellables)
     }
--- a/ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
+++ b/ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
@@ -175,6 +175,7 @@ class SendFundPresenter: SendFundPresenterProtocol {
 
     @Published private var selectedRecipient: RecipientDataType?
     @Published private var addedMemo: Memo?
+    @Published private var isUpdatingTransferCost: Bool = false
 
     var cancellables = [AnyCancellable]()
 
@@ -288,13 +289,15 @@ class SendFundPresenter: SendFundPresenterProtocol {
             .store(in: &cancellables)
 
       
-        Publishers.CombineLatest3(
+        Publishers.CombineLatest4(
             viewModel.$recipientAddress,
             viewModel.$feeMessage,
-            viewModel.$enteredAmount
+            viewModel.$enteredAmount,
+            $isUpdatingTransferCost
         )
         .receive(on: DispatchQueue.main)
-        .map { [weak self] recipientAddress, feeMessage, amount in
+        .map { [weak self] recipientAddress, feeMessage, amount, isUpdatingTransferCost in
+            guard isUpdatingTransferCost == false else { return false }
             // Called when editing the amount or address.
             if case .failure(_) = amount {
                 return false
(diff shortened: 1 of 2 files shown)


Radiokot (approved):

ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
```diff
@@ -288,13 +289,15 @@ class SendFundPresenter: SendFundPresenterProtocol {
...
             viewModel.$recipientAddress,
             viewModel.$feeMessage,
-            viewModel.$enteredAmount
+            viewModel.$enteredAmount,
+            $isUpdatingTransferCost
         )
         .receive(on: DispatchQueue.main)
-        .map { [weak self] recipientAddress, feeMessage, amount in
+        .map { [weak self] recipientAddress, feeMessage, amount, isUpdatingTransferCost in
+            guard isUpdatingTransferCost == false else { return false }
```
bisgardo, line 300: Can you please explain what exact problem this is solving? To me it just looks like it could prevent estimation to happen if I input something while the request is being slow. Meaning that this input won't be taken into account in the estimation.

ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
```diff
@@ -545,18 +554,24 @@ class SendFundPresenter: SendFundPresenterProtocol {
...
+        isUpdatingTransferCost = true
         dependencyProvider
             .transactionsService()
             .getTransferCost(transferType: transferType.actualType.toWalletProxyTransferType(), costParameters: buildTransferCostParameter())
             .sink(receiveError: { [weak self] error in
+                guard let self = self else { return }
                 Logger.error(error)
-                self?.view?.showErrorAlert(ErrorMapper.toViewError(error: error))
+                self.view?.showErrorAlert(ErrorMapper.toViewError(error: error))
+                self.isUpdatingTransferCost = false
```
bisgardo, line 565: Shouldn't you either ensure that this happens on the main thread (using `.receive(on: DispatchQueue.main)`) or make the variable atomic?

ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
```diff
@@ -545,18 +554,24 @@ class SendFundPresenter: SendFundPresenterProtocol {
     }
 
     private func updateTransferCostEstimate() {
+        isUpdatingTransferCost = true
         dependencyProvider
             .transactionsService()
             .getTransferCost(transferType: transferType.actualType.toWalletProxyTransferType(), costParameters: buildTransferCostParameter())
             .sink(receiveError: { [weak self] error in
```
bisgardo, line 561: I just realized that this is an extension on `Publisher` defined within the project. Consider adding another optional callback to run regardless of success or failure (i.e. like Java's `finally` clause). Then you could do the reset of `isUpdatingTransferCost` in a single place.

ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
ramakser: previous was done without blocking send button, and there was case when you enter recepinet address, and type amount and press send button.  in this case tx update haven't finished yet, and tx was sent with wrong amount of energy,  much lower than needed.   which cause send failed: insificient energy

bisgardo:
It's not clear to me what exact bug this is fixing. I understand that debounding the cost estimation requests makes it more likely that there's only one request in flight at a time. Was the problem before that responses might arrive out of order? If so, I think this helps but doesn't fix it. And not sending a request if there's already one in flight is wrong.

The correct solution I think is to do the debounce but always send the request. Either cancel any existing requests or ignore their response (for example pass a counter or timestamp though the pipeline and check it against the component state).

ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
bisgardo: Ok thanks. What I meant was: what problem does this exact line solve?

You also mentioned blocking the send button. Where does that happen?

ramakser:
> It's not clear to me what exact bug this is fixing. I understand that debounding the cost estimation requests makes it more likely that there's only one request in flight at a time. Was the problem before that responses might arrive out of order? If so, I think this helps but doesn't fix it. And not sending a request if there's already one in flight is wrong.
> 
> The correct solution I think is to do the debounce but always send the request. Either cancel any existing requests or ignore their response (for example pass a counter or timestamp though the pipeline and check it against the component state).

we need to disable send button till we get correct tx transaction cost, otherwise, it will fails transaction at all, like it happens now

ConcordiumWallet/Views/AccountsView/SendFunds/SendFund/SendFundPresenter.swift
ramakser: solving: outdated tx cost data, prevent sending tx while fetching new tx estimation data by disabling send button. 
where: basically here 
``` CombineLatest4(
            viewModel.$recipientAddress,
            viewModel.$feeMessage,
            viewModel.$enteredAmount,
            $isUpdatingTransferCost
        )
``` 
, this is 4 publishers that are somehow disables send button, in out case by checking  `isUpdatingTransferCost `

bisgardo: My bad, I misunderstood the context of the code. It would be helpful with a comment above the pipeline explaining that it's for updating the enabledness state of the button.

I still have one minor suggestion though ;)

```suggestion
            if isUpdatingTransferCost { return false }
```

bisgardo (approved):
I think this solution still doesn't take into account that cost estimations may return out of order.

And there probably also is the edge case that because of the debounc delay, there is a window from the last input was entered until the estimation request is sent where the button is enabled.

I'll approve as this solution is an improvement over what's there now.

ramakser:
> I think this solution still doesn't take into account that cost estimations may return out of order.
> 
> And there probably also is the edge case that because of the debounc delay, there is a window from the last input was entered until the estimation request is sent where the button is enabled.
> 
> I'll approve as this solution is an improvement over what's there now.

`there is a window from the last input was entered until the estimation request is sent where the button is enabled.` good point, will try to find solution to handle this

PR #397, opened 2024-04-08
Fix `.svg` CIS2Token images loading

ramakser:
## Purpose

Due to loading of `.svg`  CIS2Token image cause memory leak and fast app memory our crash, wee need to separate logic for regular .png/.jpeg images type and .svg

## Changes
 Add fallback for token preview load logic, in case when token image is `.svg` format


ConcordiumWallet/Views/AccountsView/AccountDetails/TokenSelectionView/CIS2TokenImagePreview.swift
```diff
@@ -0,0 +1,106 @@
...
+                case let .custom(width, height): return .init(width: width, height: height)
+            }
+        }
+    }
+    
+    let url: URL?
+    let size: CIS2TokenImagePreview.Size
+    
+    var body: some View {
+        if let url = url, url.absoluteString.contains(".svg") {
```
Radiokot, line 32: It is better to make this check case-insensitive. But anyway, this check is still not sufficient: if the image is stored in IPFS, you don't have the extension in the URL. 
Perhaps, instead, try using `WebImage` as a fallback for `AsyncImage`? So if the image can't be decoded as raster, we try to decode it as SVG as our last hope :)

Radiokot (changes requested):

Radiokot (approved):

PR #403, opened 2024-08-01
Add encryption export compliance code

ramakser:
## Purpose

To release an app we need to specify `App Encryption Export Compliance Code`  in app .plist

## Changes

Add export code for all targets


Radiokot (approved):
👏🏻

